PR #105, opened 2023-06-22
Move state store logic to journal for better separation of concerns

gvdongen:
Addresses the concerns in the feedback on PR #100 

- Logic of getting and setting state moved to the journal. The RestateContextImpl and StateMachine do not know about the local state store.
- Serialization and creation of protobuf messages is done in the RestateContextImpl
- The journal just fills the GetStateMessages with a value or empty if it finds it in the state store. 
- The journal then returns the prefilled message to the state machine and the state machine sends that to the runtime. 
- The Invocation class now just keeps the state entries in a map. No more LocalStateStore.

- During replay no get/set/clear is executed, because the state map represents the state after the replay.


slinkydeveloper:
I guess this one is no longer valid?

PR #178, opened 2023-11-13
Avoid global dyn rpc descriptor

jackkleeman:
Mutating is fine, but ensure its scoped to the base server - someone might have multiple servers created somewhere inside their application with the same names - this currently would cause a conflict


StephanEwen:
Makes sense and looks good.

+1 to merge this

PR #181, opened 2023-11-19
Add deterministic random functions

jackkleeman:


src/restate_context.ts
```diff
@@ -32,6 +32,11 @@ export interface RestateBaseContext {
    */
   serviceName: string;
 
+  /**
+   * Deterministic random methods
+   */
+  rand: Rand;
```
slinkydeveloper, line 38: I love this, will add to the Java SDK as well.

I have just one concern, I think we should be clear in the docs that this rng is not cryptographically strong, because eavesdropping the invocation id is enough to infer the generated values. You shouldn't be doing crazy stuff with this `Rand` like generating RSA keys, use the `uuid` for some private operation (e.g. assuming the invocation id is easy to reach for a MITM, would we recommend using this uuid as idempotency key in stripe?). What's your opinion on the matter?

jackkleeman: i can make it clear that deterministic -> predictable -> totally insecure for anything cryptographic. good for identifiers, or choosing between a few code paths uniformly

however, i don't think idempotency keys would be considered as something that needs security. if i have your idempotency token, i still need authenticated access to your stripe account (which would allow me to make new payments), and once i have that, I would still need to know the original parameters used (to avoid a non-determinism error of sorts from stripe), and for all that work, all i get is a `200` status code with no side effects :p

slinkydeveloper: Makes sense, choose the wording at your own discretion, let's just make it clear :)

PR #189, opened 2023-11-20
Support already completed journals in lambda handlers

### LINKED ISSUE #188: Lambda invocation seems to deadlock if it gets invoked with a journal with an OutputStreamEntry
Lambda invocation seems to deadlock if it gets invoked with a journal with an OutputStreamEntry. This might point towards a bug in the Typescript SDK. The problematic code could be

```
// Send a message back to the runtime
  send(msg: Message): Promise<void> {
  // Add the header and the body to buffer and add to the output buffer
  const msgBuffer = encodeMessage(msg);
  this.outputBuffer = Buffer.concat([this.outputBuffer, msgBuffer]);

  // An output message or suspension message is the end of a Lambda invocation
  if (
    msg.messageType === OUTPUT_STREAM_ENTRY_MESSAGE_TYPE ||
    msg.messageType === SUSPENSION_MESSAGE_TYPE
  ) {
    this.suspendedOrCompleted = true;
  }

  return RESOLVED;
}
```

jackkleeman:
Fixes #188

### DIFF
--- a/src/connection/lambda_connection.ts
+++ b/src/connection/lambda_connection.ts
@@ -23,13 +23,12 @@ const RESOLVED: Promise<void> = Promise.resolve();
 export class LambdaConnection implements Connection {
   // Empty buffer to store journal output messages
   private outputBuffer: Buffer = Buffer.alloc(0);
-  private suspendedOrCompleted = false;
 
   // Callback to resolve the invocation promise of the Lambda handler when the response is ready
   private readonly completionPromise: Promise<Buffer>;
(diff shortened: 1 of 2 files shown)


tillrohrmann (approved):
Thanks for quickly creating this PR @jackkleeman. From what I can tell, the change looks good to me.

Looking at the state machine code of the connection and also the `LambdaRestateServer`, I am wondering whether we couldn't/shouldn't add a few more unit tests for making sure that the state machine works correctly.

tillrohrmann:
fyi @igalshilman

PR #195, opened 2023-12-04
Use the new REQUIRES_ACK flag and the EntryAckMessage

slinkydeveloper:
See https://github.com/restatedev/service-protocol/issues/53

### DIFF
--- a/src/journal.ts
+++ b/src/journal.ts
@@ -202,11 +202,10 @@ export class Journal<I, O> {
       journalEntry === undefined ||
       journalEntry.messageType !== p.SIDE_EFFECT_ENTRY_MESSAGE_TYPE
     ) {
-      //TODO received completion message but for an entry that is either non existing, or is not a side effect entry
       return;
     }
 
-    // Just needs and ack without completion
+    // Just needs an ack
     journalEntry.resolve(undefined);
     this.pendingJournalEntries.delete(m.entryIndex);
   }
--- a/src/journal.ts
+++ b/src/journal.ts
@@ -18,6 +18,7 @@ import {
   CLEAR_STATE_ENTRY_MESSAGE_TYPE,
   COMPLETE_AWAKEABLE_ENTRY_MESSAGE_TYPE,
   CompletionMessage,
+  EntryAckMessage,
   GET_STATE_ENTRY_MESSAGE_TYPE,
   GetStateEntryMessage,
   INVOKE_ENTRY_MESSAGE_TYPE,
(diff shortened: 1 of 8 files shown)


src/types/types.ts
```diff
@@ -53,19 +64,21 @@ class MessageType {
...
-  static hasRequiresAckFlag(messageTypeId: bigint): boolean {
-    return this.isCustom(messageTypeId);
+  static hasRequiresAckFlag(messageType: bigint): boolean {
+    return (
+      messageType !== START_MESSAGE_TYPE &&
+      messageType !== ERROR_MESSAGE_TYPE &&
+      messageType !== SUSPENSION_MESSAGE_TYPE &&
+      messageType !== ENTRY_ACK_MESSAGE_TYPE &&
+      messageType !== COMPLETION_MESSAGE_TYPE
+    );
```
tillrohrmann, line 74: Can we say

```
messageType === SIDE_EFFECT_ENTRY_MESSAGE_TYPE
```

instead?

tillrohrmann (approved):
Thanks for adjusting the TS SDK wrt to the more explicit ack message. The changes look good to me. +1 for merging once we decide to merge https://github.com/restatedev/restate/pull/940.

src/types/types.ts
slinkydeveloper: no, the contract of this method is to return `true` if acks can be requested for this message, and this is true for any message that is an entry.

tillrohrmann: True. Thanks for the clarification.

src/journal.ts
```diff
@@ -189,14 +190,25 @@ export class Journal<I, O> {
...
       this.pendingJournalEntries.delete(m.entryIndex);
     } else {
-      if (journalEntry.messageType === p.SIDE_EFFECT_ENTRY_MESSAGE_TYPE) {
-        // Just needs and ack without completion
-        journalEntry.resolve(undefined);
-        this.pendingJournalEntries.delete(m.entryIndex);
-      } else {
-        //TODO completion message without a value/failure/empty and message is not a side effect
-      }
+      //TODO completion message without a value/failure/empty
```
slinkydeveloper, line 193: This one was just copied over from previous code.

src/journal.ts
```diff
@@ -189,14 +190,25 @@ export class Journal<I, O> {
...
+
+  public handleEntryAckMessage(m: EntryAckMessage) {
+    // Get message at that entryIndex in pendingJournalEntries
+    const journalEntry = this.pendingJournalEntries.get(m.entryIndex);
+
+    if (
+      journalEntry === undefined ||
+      journalEntry.messageType !== p.SIDE_EFFECT_ENTRY_MESSAGE_TYPE
+    ) {
+      //TODO received completion message but for an entry that is either non existing, or is not a side effect entry
```
slinkydeveloper, line 205: Doing nothing here seems sane I guess, let me remove this todo

src/journal.ts
slinkydeveloper: Looks like we should have an error here?

PR #196, opened 2023-12-05
[WIP] Saga/compensations API

### LINKED ISSUE #94: Introduce error-defer handler / sagas / compensations
In order to undo partial results in case of failures, we need a way to let the user specify an error-defer handler/compensations (a saga library). These error-defer handlers should only run if a terminal error is thrown (e.g. explicitly by the user or in the form of a `CancellationException` as part of a cancellation).

The way we register or run the error-defer handlers needs to be able to handle the situation where a `sideEffect` call is repeatedly failing and a user then cancels the invocation. In this case, we need to make sure that an associated error-defer handler is still executed because the sideEffect could be visible in the external system.

slinkydeveloper:
Draft of Saga/compensations API. Fix #94

Please look at the example for more details.

### DIFF
--- a/examples/compensations.ts
+++ b/examples/compensations.ts
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2023 - Restate Software, Inc., Restate GmbH
+ *
+ * This file is part of the Restate SDK for Node.js/TypeScript,
+ * which is released under the MIT license.
+ *
+ * You can find a copy of the license in file LICENSE in the root
+ * directory of this repository or package, or at
+ * https://github.com/restatedev/sdk-typescript/blob/main/LICENSE
+ */
+
+/* eslint-disable no-console */
+
+/*
+ * A simple example program using the Restate dynamic RPC-based API.
+ *
+ * This example primarily exists to make it simple to test the code against
+ * a running Restate instance.
+ */
+
+import * as restate from "../src/public_api";
+
+// handler implementations
+
+class MyCounter {
+  async increment(ctx: restate.RpcContext): Promise<number> {
+    const seen = (await ctx.get<number>("seen")) || 0;
+    ctx.set("seen", seen + 1);
+    // Free form compensation, can register for everything
+    ctx.registerCompensation(async () => ctx.set("seen", seen - 1));
+    return seen;
+  }
+
+  async decrement(ctx: restate.RpcContext): Promise<number> {
+    const seen = (await ctx.get<number>("seen")) || 0;
+    if (seen > 0) {
+      ctx.set("seen", seen - 1);
+      ctx.registerCompensation(async () => ctx.set("seen", seen + 1));
+    }
+    return seen;
+  }
+}
+
+class MyGreeter {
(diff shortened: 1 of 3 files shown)


examples/compensations.ts
```diff
@@ -0,0 +1,90 @@
...
+    return seen;
+  }
+}
+
+class MyGreeter {
+  async greet(ctx: restate.RpcContext, name: string): Promise<string> {
+    // Compensation on promise
+    const countSoFar = await ctx.registerPromiseCompensation(
+      ctx.rpc(counterApi).increment(name),
+      async () => ctx.send(counterApi).decrement(name)
```
slinkydeveloper, line 49: This can also get optionally another compensation function for when the promise is rejected

src/restate_context.ts
```diff
@@ -14,6 +14,18 @@ import "./utils/logger";
 import { RetrySettings } from "./utils/public_utils";
 import { Client, SendClient } from "./types/router";
 
+export interface RestatePromiseDecorator<T> {
+  withCompensation(
+    onFulfilledCompensation: (res: T) => Promise<void>,
+    onRejectedCompensation?: (err: Error) => Promise<void>
+  ): RestatePromise<T>;
```
slinkydeveloper, line 21: Down the hood this will call ctx.registerPromiseCompensation(...)

src/restate_context.ts
```diff
@@ -14,6 +14,18 @@ import "./utils/logger";
 import { RetrySettings } from "./utils/public_utils";
 import { Client, SendClient } from "./types/router";
 
+export interface RestatePromiseDecorator<T> {
+  withCompensation(
+    onFulfilledCompensation: (res: T) => Promise<void>,
+    onRejectedCompensation?: (err: Error) => Promise<void>
```
slinkydeveloper, line 20: The optional callback api looks like the [Promise.then](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then)

src/restate_context.ts
```diff
@@ -124,6 +136,13 @@ export interface RestateBaseContext {
    */
   sideEffect<T>(fn: () => Promise<T>, retryPolicy?: RetrySettings): Promise<T>;
 
+  compensatedSideEffect<T>(
+    fn: () => Promise<T>,
+    onFulfilledCompensation: (res: T) => Promise<void>,
+    onRejectedOrNotExecutedCompensation?: (err?: Error) => Promise<void>,
```
slinkydeveloper, line 142: We need this because we might execute a side effect compensation for a side effect that has not been executed, or for which we have recorded no result whatsoever.

The reason why this is fundamentally different from other methods in the context is that in the side effect case, the "external side effect" could have been applied even if the side effect is not in the journal (e.g. network error after the side effect happened, but before it got recorded), so we still need to allow users to define compensations for side effect even though they're not in the journal.

This is fundamentally similar to doing the following:

```typescript
ctx.registerCompensation(() => side_effect_compensation);
ctx.sideEffect()
```

The difference is that in the onFulfilled case we also give the result of the side effect.

PR #200, opened 2023-12-06
Ship all grpc error codes

jackkleeman:
These are useful for influencing http responses. Our users may not import grpc, and shouldn't need to know about grpc.

### DIFF
--- a/src/types/errors.ts
+++ b/src/types/errors.ts
@@ -18,26 +18,155 @@ import { JournalEntry } from "../journal";
 import { FailureWithTerminal } from "../generated/proto/javascript";
 
 export enum ErrorCodes {
+  /**
+   *  Not an error; returned on success.
+   *  HTTP 200
+   */
   OK = 0,
-  CANCELLED,
-  UNKNOWN,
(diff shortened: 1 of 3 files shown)


src/types/errors.ts
```diff
@@ -18,8 +18,25 @@ import { JournalEntry } from "../journal";
...
+  PERMISSION_DENIED,
+  RESOURCE_EXHAUSTED,
+  FAILED_PRECONDITION,
+  ABORTED,
+  OUT_OF_RANGE,
+  UNIMPLEMENTED,
+  INTERNAL,
+  UNAVAILABLE,
+  DATA_LOSS,
+  UNAUTHENTICATED,
```
slinkydeveloper, line 37: perhaps let's add the documentation you see in the gRPC status codes page here? https://github.com/grpc/grpc/blob/master/doc/statuscodes.md#status-codes-and-their-use-in-grpc

Perhaps also add the corresponding http status code. This is the mapping we use: https://connectrpc.com/docs/protocol/#error-codes

PR #203, opened 2023-12-06
Convert rpc handler responses into a JSON-safe 'any' type

### LINKED ISSUE #198: Date is being serialised incorrectly in responses
`return { date: new Date() }`

Results in:
```
{
  "response": {
    "date": {}
  }
}
```

jackkleeman:
We pass directly to the google.protobuf.Value conversion function, which is fine, except that this is not as clever as JSON.stringify, and will treat objects like Date as empty objects. To match stringify, we should respect the toJSON() function which may be present on the value to be converted.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description

Fixes #198


jackkleeman:
@igalshilman ok, I'll merge for now and file some related issues

PR #211, opened 2023-12-17
Add failure variant for GetState, PollInputStream and SleepEntryMessage

### LINKED ISSUE #210: Update SDK wrt to new service protocol definition which makes all completable journal entries fallible
Update SDK wrt restatedev/service-protocol#58.

tillrohrmann:
This commit updates the SDK to the latest service protocol version. Part of it is to enable failure variants for the GetState, PollInputStream and SleepEntryMessages so that the runtime can cancel these entries. This commit also adds tests for verifying the changes.

This fixes #210.

### DIFF
--- a/src/invocation.ts
+++ b/src/invocation.ts
@@ -14,6 +14,7 @@
 import { Message } from "./types/types";
 import { HostedGrpcServiceMethod } from "./types/grpc";
 import {
+  Failure,
   PollInputStreamEntryMessage,
   StartMessage,
 } from "./generated/proto/protocol";
@@ -37,6 +38,10 @@ enum State {
   Complete = 3,
 }
 
+type InvocationValue =
+  | { kind: "value"; value: Buffer }
+  | { kind: "failure"; failure: Failure };
+
 export class InvocationBuilder<I, O> implements RestateStreamConsumer {
   private readonly complete = new CompletablePromise<void>();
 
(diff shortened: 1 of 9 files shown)


slinkydeveloper:
We need to merge #195 first

src/invocation.ts
```diff
@@ -67,6 +72,30 @@ export class InvocationBuilder<I, O> implements RestateStreamConsumer {
...
+              };
+            } else {
+              throw new Error(
+                `PollInputStreamEntry neither contains value nor failure: ${printMessageAsJson(
+                  m
+                )}`
+              );
+            }
+          }
+
```
slinkydeveloper, line 98: Maybe isolate this in `this.handlePollInputStreamEntry` for clarity

src/state_machine.ts
```diff
@@ -201,10 +202,21 @@ export class StateMachine<I, O> implements RestateStreamConsumer {
...
+      case "value":
+        resultBytes = this.invocation.method.invoke(
+          this.restateContext,
+          this.invocation.invocationValue.value
+        );
+        break;
+      case "failure":
+        resultBytes = Promise.reject(
+          failureToError(this.invocation.invocationValue.failure, true)
+        );
```
slinkydeveloper, line 217: Maybe it's me that i don't know this code well, but it took me some time to figure out this will return terminal error. Perhaps rename the method as `failureToTerminalError` or similar.

slinkydeveloper (approved):

src/invocation.ts
tillrohrmann: Will do.

src/state_machine.ts
tillrohrmann: It only returns a terminal error if the second argument is `true`. Otherwise it will return a `RestateError`. I will create a helper method `failureToTerminalError` which wraps `failureToError`.

tillrohrmann:
Thanks for the review @igalshilman. I think the e2e tests are currently failing because the runtime image is missing some changes. Locally the e2e tests including restatedev/e2e#228 are passing.

PR #216, opened 2023-12-20
Introduce EndMessage

slinkydeveloper:
Part of https://github.com/restatedev/service-protocol/issues/49

### DIFF
--- a/src/state_machine.ts
+++ b/src/state_machine.ts
@@ -19,6 +19,8 @@ import { rlog } from "./utils/logger";
 import { clearTimeout } from "timers";
 import {
   COMPLETION_MESSAGE_TYPE,
+  END_MESSAGE_TYPE,
+  EndMessage,
   ENTRY_ACK_MESSAGE_TYPE,
   ERROR_MESSAGE_TYPE,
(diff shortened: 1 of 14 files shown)


src/state_machine.ts
```diff
@@ -303,7 +308,7 @@ export class StateMachine<I, O> implements RestateStreamConsumer {
             "Function completed with an error: " + error.message
           );
 
-          this.finishWithError(error);
+          this.sendErrorAndFinish(error);
```
tillrohrmann, line 311: Is a bit unrelated to the changes of this PR: Why is it ok to not await the result of this async function? Same question applies to the `this.finish(value)` call above.

tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. As far as I can tell, the changes look good to me. I had a question about why it is ok to not await certain calls in the `resultBytes` callback.

src/state_machine.ts
slinkydeveloper: I'm not sure why, down the stack this function is async only for waiting the `connection.end()`, that's probably why

PR #217, opened 2023-12-21
Fix side effects can be overtaken by other entries (reservations approach)

### LINKED ISSUE #197: Side effect entry can be overtaken if not awaited
The SideEffect entry writing can be overtaken if the user doesn't await on the returned promise.

To fix that, we need to internally correctly serialize the operations on the SDK state machine.

slinkydeveloper:
Fix #197


slinkydeveloper:
Closed as we don't need this anymore. See the original issue for more info.

PR #226, opened 2024-01-11
Allow printing stacktrace when invocations finish with an error

AhmedSoliman:
Allow printing stacktrace when invocations finish with an error

### DIFF
--- a/src/state_machine.ts
+++ b/src/state_machine.ts
@@ -303,9 +303,10 @@ export class StateMachine<I, O> implements RestateStreamConsumer {
           }
 
           const error = ensureError(e);
-          rlog.debugInvokeMessage(
+          rlog.trace(
             this.invocation.logPrefix,
-            "Function completed with an error: " + error.message
+            "Function completed with an error: " + error.message,
+            e
           );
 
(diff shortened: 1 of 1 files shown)


AhmedSoliman:
Worth noting that this will print the javascript stacktrace, in order to see references to the TypeScript source, the user needs to add `NODE_OPTIONS=--enable-source-maps` to env to enable this. I don't know if there is an automatic/better way to do this that doesn't require users to do anything extra.

src/state_machine.ts
```diff
@@ -303,9 +303,10 @@ export class StateMachine<I, O> implements RestateStreamConsumer {
           }
 
           const error = ensureError(e);
-          rlog.debugInvokeMessage(
+          rlog.trace(
             this.invocation.logPrefix,
-            "Function completed with an error: " + error.message
+            "Function completed with an error: " + error.message,
+            e
```
slinkydeveloper, line 309: From what I know this seems correct to me. Can you convert this PR to non-draft?

AhmedSoliman: Alrighty.

slinkydeveloper (approved):

PR #227, opened 2024-01-15
Add guard for RestateContext.sideEffect await

### LINKED ISSUE #197: Side effect entry can be overtaken if not awaited

slinkydeveloper:
Fix #197


slinkydeveloper:
TODO: Fix ctx.sleep call on retries (it should not check for the executingSideEffect guard)

Issue #235, opened 2024-01-23
Check the Typescript SDK respects the `protocol_version` flag


slinkydeveloper:
When receiving the `StartMessage`, we need to check that the `protocol_version` is compatible. This is needed to detect the user is using a compatible combination of SDK and runtime. You can check the Java version of the same task https://github.com/restatedev/sdk-java/pull/199 for details.

PR #238, opened 2024-01-25
Awakeable identifiers following the new ID scheme

AhmedSoliman:
Awakeable identifiers following the new ID scheme

Bonus: `s/assumpsions/assumptions/`

Server side in https://github.com/restatedev/restate/pull/1120 

Test Plan:

Using sample code to generate an awakeable and resolved it with
```
curl localhost:8080/dev.restate.Awakeables/Resolve -H 'content-type: application/json' -d '{"id": "prom_1NMyOAvDK2CcBjUH4Rmb7eGBp0DNNDnmsAAAAAQ", "json_result": ""}'
```

```
[restate] [2024-01-25T18:54:01.682Z] DEBUG: [CheckoutProcess/checkout] [inv_13q6r7rhzaG85fs0bfyqW8cubIUYJFjvBT] : Resuming (replaying) function.
Awakeable is:  prom_1NMyOAvDK2CcBjUH4Rmb7eGBp0DNNDnmsAAAAAQ
Awakeable COMPLETED  prom_1NMyOAvDK2CcBjUH4Rmb7eGBp0DNNDnmsAAAAAQ
```

### DIFF
--- a/src/restate_context_impl.ts
+++ b/src/restate_context_impl.ts
@@ -28,6 +28,7 @@ import {
 } from "./generated/proto/protocol";
 import {
   AWAKEABLE_ENTRY_MESSAGE_TYPE,
+  AWAKEABLE_IDENTIFIER_PREFIX,
   BACKGROUND_INVOKE_ENTRY_MESSAGE_TYPE,
   CLEAR_STATE_ENTRY_MESSAGE_TYPE,
   COMPLETE_AWAKEABLE_ENTRY_MESSAGE_TYPE,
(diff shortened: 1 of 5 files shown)


slinkydeveloper (approved):
Once it's green you can merge.

tillrohrmann (commented):
Should we bump the protocol version to mark this change as an non-backwards compatible change?

slinkydeveloper:
> Should we bump the protocol version to mark this change as an non-backwards compatible change?

There's no version to bump yet, still need to solve #235 first

AhmedSoliman:
For posterity, tests pass locally but we have a full mesh of dependency such that there is no way e2e tests will pass without merging the sdk first.

PR #250, opened 2024-02-06
Check protocol version

### LINKED ISSUE #235: Check the Typescript SDK respects the `protocol_version` flag

slinkydeveloper:
Fix #235

### DIFF
--- a/src/io/decoder.ts
+++ b/src/io/decoder.ts
@@ -32,6 +32,8 @@ type DecoderState = { state: number; header: Header | undefined; buf: Buffer };
 const WAITING_FOR_HEADER = 0;
 const WAITING_FOR_BODY = 1;
 
+export const SUPPORTED_PROTOCOL_VERSION = 0;
+
 function initalDecoderState(buf: Buffer): DecoderState {
   return {
     state: WAITING_FOR_HEADER,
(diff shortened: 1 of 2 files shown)


slinkydeveloper:
In case of protocol version incompatibility, this is the behavior:

* The runtime ends up in a retry loop, prints 

```
2024-02-08T14:04:48.582810Z WARN restate_invoker_impl
  Error when executing the invocation, retrying in 10s.
    error: [RT0007] [Unknown] unknown. For more details, look at the docs with https://docs.restate.dev/references/errors#RT0007
    restate.error.code: RT0007
    restate.invocation.id: inv_12LP4netCRaH6DrniM3ENA32Wpw4CMplKh
``` 

* The SDK prints the following error message and closes the stream:

```
[restate] [2024-02-08T14:05:08.594Z] ERROR: Error while handling connection: Error: Unsupported protocol version 1, only version 0 is supported
```

StephanEwen:
I think that is fine. Having a proper error on the runtime side would be even nicer, but this is ok for now, SDK side it prints correctly.

StephanEwen:
To have a proper error on the runtime side as well, we would need to change the following:
* In the SDK, where we get the request path, we don't immediately respond with 200, 404, etc., but we read the first chunk of bytes, like a header message and interpret that.
* If that has a protocol mismatch, we send an appropriate status code and error message
* Otherwise we start setting up the connection to read the actual Restate stream messages

PR #253, opened 2024-02-07
Add `ctx.clearAll()`

### LINKED ISSUE #251: Add `ctx.clearAll()`

slinkydeveloper:
Fix #251 

This PR also adds the method `toJSON()` in the `Message` class. This method is used by jest to print out messages in assertions.


src/local_state_store.ts
```diff
@@ -19,12 +20,14 @@ import { Empty } from "./generated/google/protobuf/empty";
 import { jsonSerialize } from "./utils/utils";
 
 export class LocalStateStore {
+  private isPartial: boolean;
   private state: Map<string, Buffer | Empty>;
 
-  constructor(readonly isPartial: boolean, state: StartMessage_StateEntry[]) {
```
StephanEwen, line 24: You could change this to `private isPartial` here and it would be the same. If you annotate constructor parameters with visibility, they become properties.

StephanEwen (approved):
Looks good. One minor comment, but +1 overall

PR #261, opened 2024-02-13
Some feedback about workflow api

slinkydeveloper:
Part of #260


StephanEwen:
I think you can now also remove the state `ALL_NAMES_STATE_NAME` and all code that maintains it, now that `clearAll()` is available.

Otherwise +1

PR #264, opened 2024-02-15
Refactor embedded handlers

igalshilman:
```ts

import express, { Request, Response } from "express";
import * as restate from "../src/public_api";

const router = restate.router({
  hello: async (ctx: restate.Context, param: { productId: string }) => {
    const response = await ctx.sideEffect(async () => {
      return (
        await fetch(`https://dummyjson.com/products/${param.productId}`)
      ).json();
    });
    return response.title;
  },
});

const routerApi: restate.ServiceApi<typeof router> = { path: "example" };

const rs = restate
  .endpoint()
  .bindRouter(routerApi.path, router)
  .connect("http://127.0.0.1:8080");

const app = express();
app.use(express.json());

app.post("/workflow", async (req: Request, res: Response) => {
  const { id } = req.body;

  const response = await rs
    .rpc(routerApi, { idempotencyKey: id, retain: 60 })
    .hello({ productId: id });
  res.send(response);
});

app.listen(3000);
```

### DIFF
--- a/examples/embedded_example.ts
+++ b/examples/embedded_example.ts
@@ -15,14 +15,12 @@ import express, { Request, Response } from "express";
 import * as restate from "../src/public_api";
 
 const router = restate.router({
-  hello: async (ctx: restate.Context, param: { productId: string }) => {
-    const response = await ctx.sideEffect(async () => {
-      return (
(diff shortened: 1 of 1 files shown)


slinkydeveloper (commented):
This looks much better!

PR #267, opened 2024-02-28
Add `src/` directory to the restate-sdk package

igalshilman:
This PR adds the `src/` directory to the list of files to be included in the `restate-sdk` package.

Running `npm pack` creates a tarball with the ./src directory included.

Note that, with this change, the bundle size is: 3,1M


StephanEwen:
looks good in general. how much of that size is generated protobuf code?

PR #269, opened 2024-03-04
Fix lambda context propagation

slinkydeveloper:
Fix issue found in verification test, where i was picking up the request id from the wrong place.


jackkleeman (approved):
lgtm!

PR #273, opened 2024-03-08
Catch send/send delayed failed promises

igalshilman:
Send/send delayed are `void` methods on our  API, but the reality is that these operations produce a promise, and this promise can actually fail. For example due to a non determinism errors. This commit makes sure to propagate errors happening in these promises back to the state machine.
See the e2e test called `test_non-determinism` in the e2e repo.

### DIFF
--- a/src/endpoint/http2_handler.ts
+++ b/src/endpoint/http2_handler.ts
@@ -48,7 +48,7 @@ export class Http2Handler {
     });
   }
 
-  private handleConnection(
+  private async handleConnection(
     url: Url,
     stream: http2.ServerHttp2Stream
   ): Promise<void> {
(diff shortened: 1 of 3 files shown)


src/endpoint/http2_handler.ts
```diff
@@ -76,7 +76,8 @@ export class Http2Handler {
       ":status": 200,
     });
     const restateStream = RestateHttp2Connection.from(stream);
-    return handleInvocation(handler, restateStream);
+    await handleInvocation(handler, restateStream);
+    return;
```
mupperton, line 80: Just a note that possibly the original implementation as it was without being an async function was to be a performance improvement?

If you can get away with directly returning a function call (which returns a promise) then it saves an extra event-loop iteration, as when you mark this function `handleConnection` as async, it has to resolve this await first, then emit another internal event that this function is now finished - I might not be explaining it very well as rushing this message, but was just a thought

PR #283, opened 2024-03-20
Expose the original Request

igalshilman:
This PR exposes the raw request on the Context. A request represents the raw request as issued for this handler. This includes the request headers, and the raw body.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -19,6 +19,30 @@ import {
 import { ContextImpl } from "./context_impl";
 
 /**
+ * Represents the original request as sent to this handler.
+ *
+ * A request object includes the request headers, and the raw unparsed
+ * request body.
+ */
+export interface Request {
+  /**
+   * The unique id that identifies the current function invocation. This id is guaranteed to be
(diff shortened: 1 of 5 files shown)


src/context.ts
```diff
@@ -18,6 +18,17 @@ import {
 } from "./types/rpc";
 import { ContextImpl } from "./context_impl";
 
+/**
+ * Represents the original request as sent to this handler.
+ *
+ * A request object includes the request headers, and the raw unparsed
+ * request body.
+ */
+export interface Request {
```
slinkydeveloper, line 27: How about moving the field `Context.id` and `Context.serviceName` (which needs to be renamed) here as well?

slinkydeveloper: i don't have `serviceName` in java sdk (seems redundant to me :shrug: ), `id` yes i have it, it's named `Request#invocationId()`: https://github.com/restatedev/sdk-java/blob/main/sdk-common/src/main/java/dev/restate/sdk/common/Request.java#L27

PR #285, opened 2024-03-20
[protocol] Use protobuf-es

igalshilman:
This is an internal change, replace [ts-proto](https://github.com/stephenh/ts-proto) with [protobuf-es](https://github.com/bufbuild/protobuf-es)

### DIFF
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -4,4 +4,4 @@ managed:
 plugins:
   - plugin: es 
     out: src/generated
-    opt: target=ts,import_extension=.ts
+    opt: target=ts,import_extension=none


buf.lock
```diff
@@ -1,2 +1,8 @@
 # Generated by buf. DO NOT EDIT.
 version: v1
+deps:
+  - remote: buf.build
+    owner: protocolbuffers
+    repository: wellknowntypes
+    commit: f17e05fe4a764a3482b8e033daec742e
+    digest: shake256:886495c560d96bc48bef49e17e3d33430fa1f6c735b0cee40ef83161824b01c2d2f342d81ca02aadeed52aa3cf4a8273d5ab9584fbb595b71e7e44240caedb00
```
slinkydeveloper, line 8: Note for myself: I think we should get rid of `google.protobuf.Empty` and define our own empty message in the protocol messages.

src/context_impl.ts
```diff
@@ -529,26 +534,32 @@ export class ContextImpl implements ObjectContext {
...
+      },
     });
   }
 
   public rejectAwakeable(id: string, reason: string): void {
     this.checkState("rejectAwakeable");
     this.completeAwakeable(id, {
-      failure: { code: UNKNOWN_ERROR_CODE, message: reason },
+      result: {
+        case: "failure",
```
slinkydeveloper, line 548: Looks much better!

slinkydeveloper (approved):

PR #287, opened 2024-03-21
Remove the retry policy in sideEffect

igalshilman:
This PR removes the SDK-side retry policy mechanism.
The runtime would still retry according to the globally defined retry police.


gvdongen:
Hi Igal, for this kind of user-facing changes. Can we file an issue in the documentation to update this?

slinkydeveloper (approved):

PR #288, opened 2024-03-21
Add context.date 

igalshilman:
This PR adds a new deterministic date to the context.

```ts
ctx.date.now() -> is the unix epoch time

ctx.date.toJSON() -> is the JSON representation of the current time 

```


gvdongen:
Hi Igal, for this kind of user-facing changes. Can we file an issue in the documentation to update this?

slinkydeveloper (approved):

PR #290, opened 2024-03-22
API iteration

igalshilman:
### This PR changes the following concepts:

1. Adds a `client` suffix for the `ctx.service` / `ctx.object`
2. simply the `endpoint` binding methods to a single `endpoint.bind()`
3. make the handlers/objects creation more explicit. 

Here is a creation/binding example:

```ts
const service = restate.service({
  name: "greeter",
  handlers: {
        greet: ..
  }
});    

export type Greeter = typeof service;

restate.endpoint().bind(service).listen()
```

And here is a call site:

```ts
import type {Greeter} from "...";

const Greeter: Greeter = { name : "greeter" };
...
await ctx.serviceClient(Greeter).greet();
...
```

Thanks @nikrooz  and @gvdongen for the feedback!

### DIFF
--- a/examples/example.ts
+++ b/examples/example.ts
@@ -11,27 +11,30 @@
 
 import * as restate from "../src/public_api";
 
-const greeter = restate.service("greeter", {
-  greet: async (ctx: restate.Context, name: string) => {
-    // blocking RPC call to a keyed service (here supplying type and path separately)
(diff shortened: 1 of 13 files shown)


slinkydeveloper (approved):
This looks great!

nikrooz:
thanks, looks great!

PR #296, opened 2024-03-25
[endpoint] Return a promise that resolves/reject on listen().

igalshilman:
This PR makes sure that the returned promise from endpoint.listen() resolves whenever the server is successfully listening on the provided port. Otherwise reject with the first error.

### DIFF
--- a/src/endpoint/endpoint_impl.ts
+++ b/src/endpoint/endpoint_impl.ts
@@ -93,7 +93,7 @@ export class EndpointImpl implements RestateEndpoint {
     return handler.handleRequest.bind(handler);
   }
 
-  listen(port?: number): Promise<void> {
+  listen(port?: number): Promise<number> {
     const actualPort = port ?? parseInt(process.env.PORT ?? "9080");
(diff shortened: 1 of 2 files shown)


src/endpoint/endpoint_impl.ts
```diff
@@ -98,9 +98,19 @@ export class EndpointImpl implements RestateEndpoint {
...
+
+    return new Promise((resolve, reject) => {
+      let failed = false;
+      server.once("error", (e) => {
+        failed = true;
+        reject(e);
+      });
+      server.listen(actualPort, () => {
+        if (!failed) {
+          resolve();
```
slinkydeveloper, line 110: How about resolving this promise with at least the current port? This is especially useful if you start listening on `0`

nikrooz: Not necessarily, I just mentioned it in case we want to be aligned with nodes api.
https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback

PR #297, opened 2024-03-26
Align side effects

igalshilman:
This PR aligns the sideEffect semantics across the SDKs.
Any exception that is thrown within the closure passed to
ctx.sideEffect is propagated to the user, and the following distinction
applies:
1. If a TerminalException was thrown, then it will get journaled before
   being rethrown.
2. If a non terminal exception is thrown, then it is _not_ journaled,
   and the current attempt terminates. This allows restate to re-invoke
   and thus retry that sideEffect.

With this commit, any terminal exception thrown from within the
sideEffect body, get's journaled

### DIFF
--- a/test/side_effect.test.ts
+++ b/test/side_effect.test.ts
@@ -131,7 +131,7 @@ describe("Greeter", () => {
     expect(result).toStrictEqual([errorMessage(f)]);
   });
 
-  it("nal exception (1) does not record the sideEffect in the journal. (2) ends the current attempt", async () => {
+  it("Local recovery from a non terminal exception, has no effect.", async () => {
     const result = await new TestDriver(new GreeterTriesToCatchNonTerminal(), [
       startMessage({ knownEntries: 1, key: "Pete" }),
       inputMessage(greetRequest("Pete")),


test/side_effect.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      startMessage({ knownEntries: 1, key: "Pete" }),
+      inputMessage(greetRequest("Pete")),
+    ]).run();
+
+    const f = failure("oh no");
+
+    expect(result).toStrictEqual([errorMessage(f)]);
+  });
+
+  it("nal exception (1) does not record the sideEffect in the journal. (2) ends the current attempt", async () => {
```
slinkydeveloper, line 134: Typo?

test/side_effect.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  it("A non terminal exception (1) does not record the sideEffect in the journal. (2) ends the current attempt", async () => {
+    const result = await new TestDriver(new GreeterThrowsRecoverable(), [
+      startMessage({ knownEntries: 1, key: "Pete" }),
+      inputMessage(greetRequest("Pete")),
+    ]).run();
+
+    const f = failure("oh no");
+
+    expect(result).toStrictEqual([errorMessage(f)]);
+  });
```
slinkydeveloper, line 132: This test is redundant right? the one you have below is enough?

slinkydeveloper (approved):

PR #306, opened 2024-04-09
[workspaces] Move to workspaces

igalshilman:


nikrooz (approved):
lgtm, just a quick remark about the TS setup. I don't think with this setup the TS likes 
```ts
import * as restate from "@restatedev/restate-sdk";
```

one way to fix this is
- Create a `tsconfig.base.json` at the root of project this will share the common ts config among all packages 
- Add `paths` for the packages you'd like use internally
```json
{
  "compilerOptions": {
    "rootDir": ".",
    "baseUrl": ".",
    "target": "esnext",
    "module": "commonjs",
    "lib": ["esnext"],
    "declaration": true,
    "declarationMap": true,
    "sourceMap": true,
    "outDir": "./dist",
    "strict": true,
    "allowJs": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "allowSyntheticDefaultImports": true,
    "paths": {
      "@restatedev/restate-sdk": [
        "./packages/restate-sdk/src/public_api.ts"
      ],
    }
  }
}
```

and the `tsconfig.json` for `restate-sdk-examples` becomes something like 
```json
{
  "extends": "../../tsconfig.base.json",
  "compilerOptions": {},
  "include": ["src/**/*.ts"]
}
```

PR #340, opened 2024-05-01
Avoid transform after invoke, which appears to break CombineablePromise

jackkleeman:
This is a hotfix, a more meaningful fix would be to implement 'transform' for combineable promises so that they stay combineable

### DIFF
--- a/packages/restate-sdk/src/context_impl.ts
+++ b/packages/restate-sdk/src/context_impl.ts
@@ -74,10 +74,9 @@ export interface CallContext {
   delay?: number;
 }
 
-export type InternalCombineablePromise<T> = CombineablePromise<T> &
-  WrappedPromise<T> & {
-    journalIndex: number;
-  };
+export type InternalCombineablePromise<T> = CombineablePromise<T> & {
+  journalIndex: number;
+};
 
 export class ContextImpl implements ObjectContext {
   // here, we capture the context information for actions on the Restate context that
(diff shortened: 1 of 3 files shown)


packages/restate-sdk/src/context_impl.ts
```diff
@@ -224,12 +224,13 @@ export class ContextImpl implements ObjectContext {
   // --- Calls, background calls, etc
 
   // DON'T make this function async!!! see sideEffect comment for details.
+  /* eslint-disable @typescript-eslint/no-explicit-any */
```
mupperton, line 227: Better to just apply this to the specific line rather than disable for the rest of the file without re-enabling?

slinkydeveloper (commented):
Can you add a test for this?

jackkleeman:
@slinkydeveloper i've added some tests for combining invokes, can confirm that they fail on main, work on this PR

packages/restate-sdk/src/context_impl.ts
```diff
@@ -667,20 +666,23 @@ export class ContextImpl implements ObjectContext {
...
+    defineProperty(p, RESTATE_CTX_SYMBOL, this);
+    defineProperty(p, "journalIndex", journalIndex);
+    defineProperty(p, "orTimeout", orTimeout.bind(this));
+
+    return p;
   }
 }
 
+// wraps defineProperty such that it informs tsc of the correct type of its output
+function defineProperty<Obj extends object, Key extends PropertyKey, T>(
```
slinkydeveloper, line 678: Good one, thanks for that!

slinkydeveloper (approved):

PR #342, opened 2024-05-02
[clients] Add an example of calling via an interface

igalshilman:


packages/restate-sdk-clients/src/api.ts
```diff
@@ -7,14 +7,14 @@ export interface Ingress {
...
-  serviceClient<P extends string, M>(
+  serviceClient<M, P extends string = string>(
     opts: ServiceDefinition<P, M>
   ): IngressClient<M>;
 
   /**
    * Create a client from a {@link VirtualObjectDefinition}.
    */
-  objectClient<P extends string, M>(
+  objectClient<M, P extends string = string>(
```
mupperton, line 17: Surely need to update the `serviceSendClient` and `objectSendClient` type signatures too?

PR #374, opened 2024-06-13
Add CF module

igalshilman:


jackkleeman (approved):
I have tested Node and Lambda and still works. Refactor looks right to me

PR #381, opened 2024-06-17
Use type-aware eslint

jackkleeman:
Mostly just type changes, I have reviewed the compiled output and there are almost no differences. What differences there are I will comment

### DIFF
--- a/packages/restate-sdk/src/endpoint/request_signing/v1.ts
+++ b/packages/restate-sdk/src/endpoint/request_signing/v1.ts
@@ -100,7 +100,9 @@ async function jwtVerify(
 
   let header: Record<string, string | undefined> = {};
   try {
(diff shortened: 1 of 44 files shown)


packages/restate-sdk/src/endpoint/request_signing/v1.ts
```diff
@@ -119,20 +121,11 @@ async function jwtVerify(
...
   }
 
-  const keyPromise = keySet.get(kid);
-  if (!keyPromise) {
+  const key = keySet.get(kid);
+  if (!key) {
     throw new Error(`kid ${header.kid} is not present in keySet`);
   }
 
-  let key: Key;
```
jackkleeman, line 127: here eslint caught a real issue; key is no longer a promise, so this await was a no op and could never throw

PR #386, opened 2024-06-18
Convert to ESM modules

jackkleeman:
This PR changes the `dist` directory to have two subdirs `esm` and `cjs` with different entrypoints in package.json
The files in the cjs directory are identical to the files previously produced in dist
The files in the esm directory are valid esm modules which will be used in esm import contexts

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "packages/restate-sdk-examples"
       ],
       "devDependencies": {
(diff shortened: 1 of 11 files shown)


nikrooz (approved):
lgtm!

jackkleeman:
@igalshilman yes it does work

jackkleeman:
and it also works with the node10 module resolution

PR #391, opened 2024-06-21
Ensure that we resolve completepromise messages on replay

jackkleeman:
Previously they would hang forever if replayed after a suspension


jackkleeman:
given we're about to release ill just wait for the e2es on the main branch

PR #399, opened 2024-06-28
Optionally typed VirtualObject state keys

mupperton:
Motivation in Discord channel: https://discord.com/channels/1128210118216007792/1255899794337959939

Due to TS strictness (good thing) this needed an unconventional default value for the Generic TState, which would be used to control whether the typings should follow the existing approach of `ctx.get<TYPE>` or should infer types from the provided TState - I was hoping to just use `null` but `null` doesn't satisfy the type `Record<string, unknown>`...

I haven't made any changes to `context_impl.ts` - not sure if needed? Due to it _not_ specifying the TState type, it uses the existing behaviour anyway from a type perspective - let me know

Usage (screenshot taken from proposal in Discord channel):
IMG_9375

### DIFF
--- a/packages/restate-sdk/src/context.ts
+++ b/packages/restate-sdk/src/context.ts
@@ -57,10 +57,13 @@ export interface Request {
   readonly body: Uint8Array;
 }
 
+type TypedState = Record<string, unknown>;
+type UntypedState = { _: never };
+
 /**
  * Key value store operations. Only keyed services have an attached key-value store.
  */
-export interface KeyValueStore {
+export interface KeyValueStore<TState extends TypedState> {
   /**
    * Get/retrieve state from the Restate runtime.
    * Note that state objects are serialized with `Buffer.from(JSON.stringify(theObject))`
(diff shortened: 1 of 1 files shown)


mupperton:
> I've noticed that some lint checks are failing

Looks like unexpected use of `any` in the `Record<string, any>`

Seems that `Record<string, unknown>` does the job too from a _quick_ test, so trying that

mupperton:
> Hi @mupperton, it turns out to be much more involved than initially thought, but still doable. If you want I can complete this based on your initial PR.

Ah yes, can see now I missed the WorkflowContext too in addition to the things you listed

Happy for you to continue the work though - I've got some other commitments at work anyway

mupperton:
> Actually it works for `type State = { .. }` but doesn't work for `interface State {}`

Yeah, I think it's because the interface doesn't extend `Record`, but using anything else, like `object` or `{ [key: string]: any }` instead of `Record` would give me a "key" type of `string | number` and sometimes `Symbol` in there too - I think I was expecting to need to cast them back to string at the time, but as I never touched the concrete implementations, maybe that will work OK and support interfaces too if desired?

mupperton:
Hey @igalshilman - thanks for this, will try out soon!

> One thing that I wasn't able to do is to provide a type parameter for the entire object, like so:
> 
> ```
> object<MyState>({ ... })
> ```

Yeah I think this is probably too much of an ask as you say, but also because any handler defined in another file from the object itself will never have any awareness of the object without some re-importing of the object's type etc which could get messy, but the current solution is more than enough I believe for those that want to opt-in to this for now

PR #400, opened 2024-07-03
Allow providing a custom logger

### LINKED ISSUE #339: Allow custom loggers
Allow users to register their own logger, for example to support JSON logging, or a different transport than the console/stdout such as to a file or to a remote location

This would be optional and if no logger provided then logs stay as-is

This should be used for SDK logs and user logs

Possible solution is that users register a callback function that receives an object of named args that they can then call their own logger with

It should allow all existing logged fields to be exposed, but it would also be nice if there was an extra field called `userArgs` or similar that is perhaps an array of the args the ctx logger was called with, e.g.
```ts
ctx.console.info('A', 'B', {})
```
`userArgs` = `['A', 'B', {}]`
Or if you want to reserve the first arg as a field called `message` and then any additional args are provided as `userArgs` or `extraArgs` or something like that, also great

A field to disambiguate SDK and user logs would also be appreciated e.g. `sdk: true` or `source: 'sdk' // or 'user'`

jackkleeman:
Fixes #339

Allows you to provide a logging function that receives some parameters about the log (the context of the invocation, the level, where it came from) in addition to the message and extra params and allows the user to decide what to do with it

### DIFF
--- a/packages/restate-sdk/src/endpoint.ts
+++ b/packages/restate-sdk/src/endpoint.ts
@@ -15,6 +15,7 @@ import type {
   ServiceDefinition,
   WorkflowDefinition,
 } from "@restatedev/restate-sdk-core";
+import type { Logger } from "./logger.js";
 
 /**
  * Utility interface for a bundle of one or more services belonging together
@@ -56,6 +57,29 @@ export interface RestateEndpointBase<E> {
    *
    */
   withIdentityV1(...keys: string[]): E;
+
+  /**
+   * Replace the default console-based {@link Logger}
+   * @param logger
+   * @example
+   * Using console:
+   * ```ts
+   *     restate.setLogger((params, message, ...o) => {console.log(`${params.level}: `, message, ...o)})
+   *  ```
+   * @example
+   * Using winston:
+   * ```ts
+   *     const logger = createLogger({ ... })
+   *     restate.setLogger((params, message, ...o) => {logger.log(params.level, {invocationId: params.context?.invocationId}, [message, ...o].join(' '))})
+   *  ```
+   * @example
+   * Using pino:
+   * ```ts
+   *     const logger = pino()
+   *     restate.setLogger((params, message, ...o) => {logger[params.level]({invocationId: params.context?.invocationId}, [message, ...o].join(' '))})
+   *  ```
+   */
+  setLogger(logger: Logger): void;
 }
 
 /**
--- a/packages/restate-sdk/src/endpoint.ts
+++ b/packages/restate-sdk/src/endpoint.ts
@@ -79,7 +79,7 @@ export interface RestateEndpointBase<E> {
    *     restate.setLogger((params, message, ...o) => {logger[params.level]({invocationId: params.context?.invocationId}, [message, ...o].join(' '))})
    *  ```
    */
-  setLogger(logger: Logger): void;
+  setLogger(logger: Logger): E;
 }
 
 /**
--- a/packages/restate-sdk/src/logger.ts
+++ b/packages/restate-sdk/src/logger.ts
@@ -60,6 +60,7 @@ function logLevelFromName(name?: string): RestateLogLevel | null {
 export type LogParams = {
   source: LogSource;
   level: RestateLogLevel;
+  replaying: boolean;
   context?: LoggerContext;
 };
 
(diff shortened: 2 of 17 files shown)


packages/restate-sdk/src/logger.ts
```diff
@@ -56,21 +57,57 @@ function logLevelFromName(name?: string): RestateLogLevel | null {
...
+  source: LogSource;
+  level: RestateLogLevel;
+  context?: LoggerContext;
+};
+
+export type Logger = (
+  params: LogParams,
+  message?: any,
+  ...optionalParams: any[]
+) => void;
```
slinkydeveloper, line 70: I guess this would not work ootb with winston and friends without writing some adapter code?

slinkydeveloper: Perhaps let's have an example with `winston`.

packages/restate-sdk/src/logger.ts
```diff
@@ -56,21 +57,57 @@ function logLevelFromName(name?: string): RestateLogLevel | null {
...
+ * Replace the default console-based {@link Logger}
+ * @param newLogger
+ * @example
+ * ```ts
+ *     restate.setLogger((params, message, ...o) => {console.log(`${params.level}: `, message, ...o)})
+ *  ```
+ */
+export function setLogger(newLogger: Logger) {
+  logger = newLogger;
 }
```
slinkydeveloper, line 111: Would it be better if this is an argument of the endpoint?

`restate.endpoint().usingLogger(Logger)` etc

slinkydeveloper (commented):
Just quickly skimmed over it.

I wonder if we should do a 1.1 release for this, as it looks to me like a new feature and for new features we should do minor releases I guess.

packages/restate-sdk/src/logger.ts
jackkleeman: they need to write some code that converts it into winston, its not oob, but i think thats ok. ill write an example

slinkydeveloper: Is there some sort of "standard-ish" interface for those loggers in nodejs we can try our interface to match? Perhaps you can look at winston and pino as two examples? If it's not possible to come up with a unique interface, then no big deal.

packages/restate-sdk/src/logger.ts
```diff
@@ -113,52 +150,41 @@ function formatLogPrefix(context?: LoggerContext): string {
...
-  },
-};
-
 function loggerForLevel(
+  source: LogSource,
   level: RestateLogLevel,
   shouldLog: () => boolean,
-  prefix: string
+  context?: LoggerContext
 ): PropertyDescriptor {
```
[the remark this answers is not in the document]
jackkleeman: if someone provides a custom logger eg winston, the thing that matters is the log level of winston. we should not drop messages before they get to their logger impl (except when replaying)

packages/restate-sdk/src/endpoint.ts
```diff
@@ -56,6 +57,29 @@ export interface RestateEndpointBase<E> {
...
+   *     restate.setLogger((params, message, ...o) => {logger.log(params.level, {invocationId: params.context?.invocationId}, [message, ...o].join(' '))})
+   *  ```
+   * @example
+   * Using pino:
+   * ```ts
+   *     const logger = pino()
+   *     restate.setLogger((params, message, ...o) => {logger[params.level]({invocationId: params.context?.invocationId}, [message, ...o].join(' '))})
+   *  ```
+   */
+  setLogger(logger: Logger): void;
```
mupperton, line 82: Return type is wrong, prevents chaining

packages/restate-sdk/src/logger.ts
```diff
@@ -56,22 +57,49 @@ function logLevelFromName(name?: string): RestateLogLevel | null {
...
 
-function logFunction(level: RestateLogLevel) {
-  switch (level) {
+export type LogParams = {
+  source: LogSource;
+  level: RestateLogLevel;
+  context?: LoggerContext;
+};
+
+export type Logger = (
```
mupperton, line 66: Please can this be exposed on public_api so I can define a typed logging callback in a different file to where I use `endpoint().setLogger(customRestateLogger)` otherwise I don't have the types for the args

jackkleeman: yep

packages/restate-sdk/src/logger.ts
```diff
@@ -113,58 +139,72 @@ function formatLogPrefix(context?: LoggerContext): string {
...
-  const name = logLevelName(level);
-  const fn = logFunction(level);
-
   return {
-    get: () => {
+    get: (): Logger => {
       if (!shouldLog()) {
         return () => {
           // empty logger
         };
```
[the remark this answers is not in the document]
jackkleeman: we could always send but with a flag and let them filter - wdyt @mupperton ?

slinkydeveloper (approved):

PR #409, opened 2024-08-02
Make it possible to get extra args provided to functional handlers

jackkleeman:
https://discord.com/channels/1128210118216007792/1136316066696532069/1268606950833127454

### DIFF
--- a/packages/restate-sdk/src/context.ts
+++ b/packages/restate-sdk/src/context.ts
@@ -62,6 +62,8 @@ export interface Request {
    * Cloudflare workers: [Env, ExecutionContext]
    * Deno: [ConnInfo]
    * Bun: [Server]
+   * These arguments can contain request-specific values that could change after a suspension.
+   * Care should be taken to use them deterministically.
    */
   readonly extraArgs: unknown[];
 }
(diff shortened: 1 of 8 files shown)


packages/restate-sdk/src/context.ts
```diff
@@ -55,6 +55,15 @@ export interface Request {
...
   readonly body: Uint8Array;
+
+  /**
+   * Extra arguments provided to the request handler:
+   * Lambda: [Context]
+   * Cloudflare workers: [Env, ExecutionContext]
+   * Deno: [ConnInfo]
+   * Bun: [Server]
+   */
+  readonly extraArgs: unknown[];
```
slinkydeveloper, line 66: There's probably something to say here about determinism. People should be extra careful about those arguments, and really consume them only within `ctx.run`.

Perhaps we should rather think about exposing them only within the `run` closure (e.g. as optional argument of the run closure itself.)

jackkleeman: agree, although it applies to many fields in this struct (eg body, attemptHeaders)

slinkydeveloper: `body` is deterministic, `headers` too. `attemptHeaders` though is indeed not deterministic...

PR #416, opened 2024-08-12
Introduce Serde

igalshilman:
This PR adds the concept of a `Serde`.

### RPC

With this, it is possible to configure both the handlers and the call sites to use any customized serialization.

Here is an example of an handler that would like to accept binary data:

```ts
{
    greeter: restate.handlers.object.exclusive(
      {
        input: restate.serde.binary,
        output: restate.serde.binary,
      },
      async (ctx: restate.ObjectContext, data: Uint8Array) => {
         console.log("Received binary data", data);
        return data;
      }
    ),
}
```

To call that handler with a raw input (instead of a default JSON) one can do:

```ts
ctx.objectClient(Foo).greeter(buf,  restate.rpc.opts({ 
input: restate.serde.binary,
output: restate.serde.binary})
```

It is also possible to provide your own implemention of the `Serde<T>` interface.

### State

Serde can be provided also to state access:

```ts
const buf = await ctx.get("foo", restate.serde.binary) 
```

### Awekables/ Promises

```ts
ctx.resolveAwekable(buf, restate.serde.binary)
```


packages/restate-sdk/src/types/rpc.ts
```diff
@@ -35,26 +35,101 @@ import { defaultSerde, type Serde } from "./serde.js";
...
+    return this.opts;
+  }
+}
+
+export type ClientSendOptions<I> = {
+  input?: Serde<I>;
+  delay?: number;
+};
+
+export class SendOpts<I> {
```
marcus-sa, line 67: What's the point of using classes when you already have interfaces for the options?
Seems completely unnecessary and bad developer experience

marcus-sa: @igalshilman I see. That makes sense!
I'm looking forward to it.
That would greatly simplify my usage and provide type safety 
https://github.com/marcus-sa/deepkit-restate/blob/16c94bf108476ae78325f3f33cad02f8adf71273/src/restate-server.ts#L204

PR #417, opened 2024-08-26
Add support for Protocol version V2

slinkydeveloper:
Although we don't implement any feature from Protocol version 2, we now solve the issue of the discovery, where the sdk declared to support v2 and then failing on the invocation path.


slinkydeveloper:
This solution doesn't work with back-compat

PR #438, opened 2024-10-01
Use request identity verification from shared core

### LINKED ISSUE #436: Deno 2.0 breaks request identity validation
It appears that Deno 2.0 now sets the process global and pretends to be node, which means we won't use webcrypto in deno. Sadly their non webcrypto node polyfill doesnt support ed25519, even though the webcrypto does. This may not matter if we are moving to the shared core

slinkydeveloper:
Fix #436


jackkleeman (approved):
it gives me great relief to see my handrolled jwt decoder impl removed :)

PR #455, opened 2024-10-30
Add key to LoggerContext

### LINKED ISSUE #453: Include ctx.key in `ctx.console.*` output
Currently the output from `ctx.console.log` and similar includes the service name, the invocation ID, and the timestamp. It would be very useful IMO if it included `ctx.key` for objects and workflows as well so we don't have to manually include it in every call. Any thoughts on this?

slinkydeveloper:
Fix #453

### DIFF
--- a/packages/restate-sdk/src/logger.ts
+++ b/packages/restate-sdk/src/logger.ts
@@ -115,18 +115,19 @@ function readRestateLogLevel(): RestateLogLevel {
 }
 
 export class LoggerContext {
-  readonly fqMethodName: string;
+  readonly invocationTarget: string;
 
   constructor(
     readonly invocationId: string,
-    packageName: string,
-    serviceName: string,
-    handlerName: string,
+    readonly serviceName: string,
+    readonly handlerName: string,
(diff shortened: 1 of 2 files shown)


packages/restate-sdk/src/logger.ts
```diff
@@ -115,26 +115,27 @@ function readRestateLogLevel(): RestateLogLevel {
...
+    readonly serviceName: string,
+    readonly handlerName: string,
+    readonly key?: string,
     readonly additionalContext?: { [name: string]: string }
   ) {
-    this.fqMethodName = packageName
-      ? `${packageName}.${serviceName}/${handlerName}`
-      : `${serviceName}/${handlerName}`;
+    this.invocationTarget =
+      key === undefined || key.length === 0
```
mupperton, line 128: Just curious, is there an actual scenario where the key is `""` (empty string) or is this just defensive coding?

slinkydeveloper: defensive coding. IIRC restate-server disallows empty keys.

PR #460, opened 2024-11-01
Add @restate/restate-sdk-testcontainers

jackkleeman:


packages/restate-sdk-testcontainers/src/restate_test_environment.ts
```diff
@@ -0,0 +1,340 @@
...
+  return restateHttpServer;
+}
+
+// Prepare the restate testcontainer
+async function prepareRestateTestContainer(
+  restateServerPort: number,
+  mutateTestContainer: (c: GenericContainer) => GenericContainer = (c) => c
+): Promise<StartedTestContainer> {
+  const restateContainer = new GenericContainer(
+    "docker.io/restatedev/restate:1.1"
```
slinkydeveloper, line 56: Can we make this configurable?

slinkydeveloper: Ah ok nvm should be configurable through the mutateTestContainers I guess...

packages/restate-sdk-testcontainers/src/restate_test_environment.ts
```diff
@@ -0,0 +1,340 @@
...
+    );
+
+  // This MUST be executed before starting the restate container
+  // Expose host port to access the restate server
+  await TestContainers.exposeHostPorts(restateServerPort);
+
+  // Start restate container
+  const startedRestateContainer = await mutateTestContainer(
+    restateContainer
+  ).start();
```
slinkydeveloper, line 75: Actually how about the following: remove this `mutateTestContainer` function and replace it with `restateContainerFactory` that just creates the restate container and can inject all the configuration, including image name. And by default the factory just does the same thing as line 55-57.

packages/restate-sdk-testcontainers/src/restate_test_environment.ts
```diff
@@ -0,0 +1,340 @@
...
+    // eslint-disable-next-line @typescript-eslint/await-thenable
+    const table = (await tableFromIPC(res.body)).toArray() as {
+      key: string;
+      value: Uint8Array;
+    }[];
+
+    return table;
+  }
+
+  public async set<TValue, TKey extends keyof TState = string>(
```
slinkydeveloper, line 258: I think this and the below function deserve a very big doc comment saying that setting the state is an asynchronous process and after the await, there is not even a guarantee that this actually happened.

IDK perhaps i would even consider having this in the name somehow, like `setAsync` :shrug:

slinkydeveloper (commented):
just few comments, lgtm

slinkydeveloper:
I think to finish this we should have a mention about that in the documentation https://github.com/restatedev/documentation/issues/472

PR #471, opened 2024-12-20
Add `name` property to custom error classes

rashfael:
Adding the `name` property to custom errors would help us with serializing errors and re-throwing TerminalErrors through user-defined code sandboxes. Otherwise `name` on these custom errors is just `Error`.

### DIFF
--- a/packages/restate-sdk/src/types/errors.ts
+++ b/packages/restate-sdk/src/types/errors.ts
@@ -41,6 +41,7 @@ export function ensureError(e: unknown): Error {
 
 export class RestateError extends Error {
   public readonly code: number;
+  public name = "RestateError";
 
   constructor(message: string, options?: { errorCode?: number; cause?: any }) {
     super(message, { cause: options?.cause });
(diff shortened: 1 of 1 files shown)


rashfael:
We're running user code and no-code runners as wasm components that can call apis on the host. For example if we throw a TerminalError inside our database update api the call stack looks like this:

- Service / gost
- wasm guest
- Host API

Every time we cross the boundary to wasm (so if we call wasm or wasm calls the api) we are serializing/deserializing the payload with json, including errors. If the host api throws an error, it needs to pass through the boundary twice.

When serializing we loose the js prototype chain and `instanceof` doesn't work anymore.
On the service side we re-instantiate TerminalErrors whenever we can so the restate sdk can check `instanceof` again.

Currently we try to serialize `error.constructor.name`, which is a bit hacky though (and minifiers are usually not preserving constructor names without being told).

Setting `name` would help us and be in line with built-in errors. `(new TypeError()).name` for example is `TypeError`.

PR #490, opened 2025-02-27
Add /health endpoint

### LINKED ISSUE #489: Do we have a health check path for restate services?
Hello, I tried to deploy the restate services to our internal platform, but it seems that the restate service doesn't have a health check endpoint/path which is required in our internal platform.

When I try to curl the possible health check endpoint for restate service, the output as below

```
➜   curl -v  http://localhost:9004/health 
*   Trying [::1]:9004...
* Connected to localhost (::1) port 9004
> GET /health HTTP/1.1
> Host: localhost:9004
> User-Agent: curl/8.4.0
> Accept: */*
> 
* Received HTTP/0.9 when not allowed
* Closing connection
curl: (1) Received HTTP/0.9 when not allowed
```

do we have a health check path for restate services that support http1.1, am I missing something here, or if there is any way to workaround, thx

slinkydeveloper:
Fix #489

### DIFF
--- a/packages/restate-sdk/src/endpoint/handlers/generic.ts
+++ b/packages/restate-sdk/src/endpoint/handlers/generic.ts
@@ -137,11 +137,21 @@ export class GenericHandler implements RestateHandler {
     }
 
     if (parsed.type === "unknown") {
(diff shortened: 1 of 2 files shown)


wu-json:
Hello - thanks for adding this! Was wondering when this would make it into a release.

PR #496, opened 2025-03-10
Expose invocation id of a call

igalshilman:


packages/restate-sdk/src/context_impl.ts
```diff
@@ -120,6 +125,29 @@ export class ContextImpl implements ObjectContext, WorkflowContext {
...
+    invocationId: InvocationId,
+    serde?: Serde<T>
+  ): CombineablePromise<T> {
+    return this.processCompletableEntry(
+      (vm) => vm.sys_attach_invocation(invocationId),
+      completeUsing(SuccessWithSerde(serde ?? defaultSerde()), Failure)
+    );
+  }
+
+  getOutput<T>(invocationId: InvocationId): CombineablePromise<T> {
```
slinkydeveloper, line 144: This is missing serde optional parameter as above.

packages/restate-sdk/src/context_impl.ts
```diff
@@ -120,6 +125,29 @@ export class ContextImpl implements ObjectContext, WorkflowContext {
...
+    return this.processCompletableEntry(
+      (vm) => vm.sys_get_invocation_output(invocationId),
+      completeUsing(SuccessWithSerde(defaultSerde()), Failure)
```
slinkydeveloper, line 147: This is incorrect, you need to handle the case with `Empty`. When get invocation output returns empty, the invocation did not complete yet (there's the same API in the ingress client)

slinkydeveloper (commented):
Can you rebase on main and fix the TODOs in the test services? Now you can enable few new tests.

packages/restate-sdk/src/promises.ts
```diff
@@ -278,6 +300,19 @@ export class RestatePendingPromise<T> implements RestatePromise<T> {
   readonly [Symbol.toStringTag] = "RestatePendingPromise";
 }
 
+export class InvocationPendingPromise<T>
+  extends RestatePendingPromise<T>
+  implements InvocationPromise<T>
```
slinkydeveloper, line 305: No point in an additional class, just make `RestatePendingPromise` implement `InvocationPromise`

packages/restate-sdk/src/context_impl.ts
```diff
@@ -188,17 +210,32 @@ export class ContextImpl implements ObjectContext, WorkflowContext {
...
 
-      // TODO eventually we return this promise back to the user
-      // const invocationIdPromise = this.createInvocationIdPromise(
-      //   call_handles.invocation_id_completion_id
-      // );
+      const invocationIdHandle = call_handles.invocation_id_completion_id;
+      const invocationHandle = () => {
+        // create a promise, on demand that resolves to an invocation id if needed.
+        return this.createInvocationIdPromise(invocationIdHandle);
+      };
```
slinkydeveloper, line 217: This is wrong, because it has the side effect that once you await the invocation id of a call once, you cannot await it again. I think it's better to create the promise outside the closure, so that the user in case it calls twice invocationId(), it gets the same promise.

```suggestion
      const p = this.createInvocationIdPromise(invocationIdHandle)
      const invocationHandle = () => {
        // create a promise, on demand that resolves to an invocation id if needed.
        return p;
      };
```

slinkydeveloper: > how about caching it?

This works, as long as you give to the user the promise you created the first time.

packages/restate-e2e-services/src/test_utils.ts
```diff
@@ -71,7 +71,7 @@ const o = restate.service({
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       invocationId: string
     ): Promise<void> {
-      // TODO add cancel invocation here!
+      ctx.cancel(restate.InvocationIdParser.fromString(invocationId));
       return Promise.reject(new TerminalError("cancel is not supported yet"));
```
slinkydeveloper, line 75: Probably you need to remove this :)
```suggestion
```

PR #502, opened 2025-03-12
Add zod module

igalshilman:
```ts
const User = z.object({
  name: z.string(),
  age: z.number().optional(),
});

type User = z.infer<typeof User>;

const greeter = restate.service({
  name: "greeter",
  handlers: {
    greet: restate.handlers.handler(
      { input: serde.zod(User, "user") },
      async (ctx: restate.Context, user: User) => {
        return `Hello ${user.name}`;
      }
    ),
  },
});
```

### DIFF
--- a/packages/restate-sdk-zod/package.json
+++ b/packages/restate-sdk-zod/package.json
@@ -1,0 +1,64 @@
+{
+  "name": "@restatedev/restate-sdk-zod",
+  "version": "1.4.0",
+  "description": "Restate Typescript SDK Zod",
+  "author": "Restate Developers",
+  "license": "MIT",
+  "email": "code@restate.dev",
+  "homepage": "https://github.com/restatedev/sdk-typescript#readme",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/restatedev/sdk-typescript.git"
+  },
+  "bugs": {
+    "url": "https://github.com/restatedev/sdk-typescript/issues"
+  },
+  "type": "module",
+  "sideEffects": false,
+  "main": "./dist/cjs/src/public_api.js",
+  "types": "./dist/cjs/src/public_api.d.ts",
+  "module": "./dist/esm/src/public_api.js",
+  "exports": {
+    ".": {
+      "import": {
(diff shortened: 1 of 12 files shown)


packages/restate-sdk-zod/package.json
```diff
@@ -0,0 +1,64 @@
...
+      "require": {
+        "types": "./dist/cjs/src/public_api.d.ts",
+        "default": "./dist/cjs/src/public_api.js"
+      }
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "devDependencies": {
```
nikrooz, line 36: I think this should be
```suggestion
  "dependencies": {
```

packages/restate-sdk-zod/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "@restatedev/restate-sdk-core": "^1.4.0"
+  },
+  "dependencies": {
```
nikrooz, line 39: and maybe this should be
```suggestion
  "peerDependencies": {
```

nikrooz (approved):
lgtm 🤩

marcus-sa:
I appreciate that this is being made optional and not mandatory.

mupperton:
@igalshilman - What am I doing wrong?

image

image

With deps:
```json
    "@restatedev/restate-sdk": "0.0.0-SNAPSHOT-20241220152906",
    "@restatedev/restate-sdk-zod": "0.0.0-SNAPSHOT-20250312193516",
    "zod": "3.23.8",
```

mupperton:
@igalshilman Found my issue, I had `zod@3.23.8` but `zod-to-json-schema` required `^3.24.1` so had 2 versions installed with differing zod types

Btw that example doesn't have the zod packages listed in the package dependencies

mupperton:
Ah, the @latest tag on npm is pointing to an older snapshot, that's why I needed 2 args, but yeah the @dev tag does not

For specifying the deps, my only suggestion would be aligning the minimum zod version with that of `zod-to-json-schema`'s minimum version, although it may be a pain to maintain if you bump versions on either

PR #512, opened 2025-03-26
Empty input still parsed by zod in ZodSerde

### LINKED ISSUE #510: zod serde issue with undefined input on handler
If a service handler is using the zod serde with say for example a simple string validator: `const inputSchema = z.string()` then the type of the input argument on the handler is inferred correctly as `string`

But if you invoke the endpoint with no input, the input arg at runtime is `undefined` which is obviously wrong

I believe this is because an empty input bypasses the zod validation in the serde deserialize method, and should be fixed like so:
```ts
  deserialize(data: Uint8Array): T {
    // if (data.length === 0) {
    //   return undefined as any;
    // }
    // const js = JSON.parse(new TextDecoder().decode(data));

    const js = data.length === 0 ? undefined : JSON.parse(new TextDecoder().decode(data));

    const res = this.schema.safeParse(js);
    if (res.success) {
      return res.data;
    }
    throw res.error;
  }
```

Schemas such as `z.string().optional()` which already correct infer the input type as `string | undefined` will also work as expected

mupperton:
Fixes #510

### DIFF
--- a/packages/restate-sdk-zod/src/serde_api.ts
+++ b/packages/restate-sdk-zod/src/serde_api.ts
@@ -36,10 +36,10 @@ class ZodSerde<T extends z.ZodType> implements Serde<z.infer<T>> {
   }
 
   deserialize(data: Uint8Array): T {
-    if (data.length === 0) {
-      return undefined as any;
-    }
-    const js = JSON.parse(new TextDecoder().decode(data));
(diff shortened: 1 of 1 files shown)


mupperton:
Ah the joys of manually making changes in the web UI 😄 

Should be all good now

PR #533, opened 2025-04-30
Fix tests

nikrooz:


packages/restate-sdk-examples/package.json
```diff
@@ -41,8 +41,7 @@
   },
   "devDependencies": {
     "tsx": "^4.15.7",
-    "@restatedev/restate-sdk-testcontainers": "^1.5.4",
-    "testcontainers": "^10.24.1"
```
nikrooz, line 45: When ⁠`GenericContainer` was imported from two separate modules (⁠restate-sdk-examples and ⁠restate-sdk-testcontainers), it created two distinct versions of the class—causing issues with testcontainers. To resolve this, I introduced ⁠`RestateGenericContainer` as a single, unified reference.

packages/restate-sdk/test/service_bind.test.ts
```diff
@@ -30,10 +30,8 @@ const greeterFoo = restate.service({
 
 describe("BindService", () => {
   it("should preserve `this`", async () => {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access,@typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-assignment
-    const handler = greeterFoo["service"]["greet"];
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-    expect(await handler({}, "abc")).toEqual("abc");
+    // @ts-ignore
+    expect(await greeterFoo.service.greet({}, "abc")).toEqual("abc");
```
nikrooz, line 34: The value of `this` depends how a function is called rather than how it was created :) 
extracting the method and calling it directly makes `this` `undefined`.

packages/restate-sdk-testcontainers/src/restate_test_environment.ts
```diff
@@ -167,6 +167,11 @@ export class RestateTestEnvironment {
 export type TypedState = Record<string, any>;
 export type UntypedState = { _: never };
 
+export class RestateGenericContainer extends GenericContainer {
```
slinkydeveloper, line 170: My understanding here is that you can't get away without adding a new public API?

If not, then name this `RestateContainer`

nikrooz: unfortunately not. done ✅

slinkydeveloper (approved):

PR #539, opened 2025-05-13
Custom RPC opts mapper

mupperton:
Motivation:
I want to be able to set a HTTP header on every RPC call, which I currently achieve by iterating over my handlers and wrapping them in a new function which modifies the `ctx` object to overwrite the RPC call methods and auto-inject the HTTP header into an existing ClientCallOptions object or create a new one, on each call - a bit of a hassle

Possible other use cases:
- Set a different serde by default for RPC ops
- logging
- any RPC hooks of interest

I considered other approaches to this problem, such as allowing it to be set as handler metadata, which would then become a property of the HandlerWrapper class, but this didn't feel logical to belong to an individual handler, and felt more global - of course the mapper function can be coded to opt-out of default behaviour if a handler actually provides some Opts or SendOpts

Happy to hear thoughts on this, and please feel free to enhance the JSDoc on the public API if not sufficient

Happy to shorten the property names (e.g. `clientCallOptsMapper`) if too verbose, but I struggled to think of a good short name

### DIFF
--- a/packages/restate-sdk/src/endpoint/handlers/generic.ts
+++ b/packages/restate-sdk/src/endpoint/handlers/generic.ts
@@ -304,10 +304,12 @@ export class GenericHandler implements RestateHandler {
       attemptCompletedSignal: abortSignal,
     };
 
+    const handlerComponent = handler.component();
+
     // Prepare logger
     const loggerContext = new LoggerContext(
       input.invocation_id,
-      handler.component().name(),
+      handlerComponent.name(),
       handler.name(),
       handler.kind() === HandlerKind.SERVICE ? undefined : input.key,
       invocationRequest,
@@ -355,7 +357,9 @@ export class GenericHandler implements RestateHandler {
       invocationRequest,
       invocationEndPromise,
       inputReader,
-      outputWriter
+      outputWriter,
+      handlerComponent.clientCallOptsMapper,
+      handlerComponent.clientSendOptsMapper
     );
 
     // Finally invoke user handler
--- a/packages/restate-sdk/src/common_api.ts
+++ b/packages/restate-sdk/src/common_api.ts
@@ -32,6 +32,8 @@ export type {
   SendClient,
   ClientCallOptions,
   ClientSendOptions,
+  ClientCallOptsMapper,
+  ClientSendOptsMapper,
   RemoveVoidArgument,
 } from "./types/rpc.js";
 export {
--- a/packages/restate-sdk/src/context_impl.ts
+++ b/packages/restate-sdk/src/context_impl.ts
@@ -38,7 +38,12 @@ import {
   TerminalError,
   UNKNOWN_ERROR_CODE,
 } from "./types/errors.js";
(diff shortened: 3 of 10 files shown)


packages/restate-sdk/src/endpoint/handlers/generic.ts
```diff
@@ -304,10 +304,12 @@ export class GenericHandler implements RestateHandler {
       attemptCompletedSignal: abortSignal,
     };
 
+    const handlerComponent = handler.component();
```
mupperton, line 307: In hindsight, this extraction is not necessary, can inline all 3 usages if preferred

mupperton:
> * One thing that you'd need to be aware of is that the `input` / `output` keys are used at discovery and they report the shape of the input and the output. I'm not sure how clear it is that these new `endpoint` options are not effecting this.

This isn't touching the serdes of the handlers, so no issues with discovery, this is about when doing an RPC call

> * Also, I'm wondering rather this is a bit too specific, as probably what is a bit more general than this is e is some sort of a request middle-ware/interceptor.

Yeah that's fair, this could instead be done as an interceptor on `genericCall` and `genericSend`, would that be better?

> * Should this be configured per service as well, not only per endpoint? I can imagine that some services might have different requirements (different header's to set?)

Potentially yes, although I'd imagine if you had a specific header for a specific use-case you would want to add that in-line in usages, this was meant more as a global setting, like you always use binary data in every handler, or always use your own custom JSON serde, so you don't want to overwrite the input/output serde of every RPC call/send in addition to the handler serde changes made for discovery

I think possibly the interceptor style approach on genericCall/genericSend might be the way to go though instead, and hopefully a cleaner implementation - let me know if you agree

slinkydeveloper:
Hey @mupperton, I've read through this proposal, here's my 2 cents.

I generally agree with the need, but I wonder if we can perhaps serve this API in "stages", just focusing on the set of defaults config options you need for all the requests and without starting with an "endpoint global" configuration. 

So how about this to begin with:

```
val clientFactory = ctx.clientFactory(opts?: { /** all the default opts for all the RPCs here **/})
```

This `clientFactory` basically contains `genericCall` and `genericSend`, and the various `workflowSend`/`workflowClient`/`serviceSend`/`serviceClient` etc etc (we could even shorter the names to just `service`/`workflow`/`object` and the send variants. Ofc if the user provides additional _options_ on the callsite, we need to merge them.

Then from this, we could introduce a "service options scoped" client factory (e.g. in the new service `options`). Then after that, we can go and put it on the endpoint if we need to... Tell me if that makes sense to you!

Re: using this mapper for logging, i think that would be a bad usage of this abstraction tbh, and i would do that in a different way (maybe an opt in the rpc opts, idk).

Generally the mapper idea (as opposed to "static" config options) can work for me, but I wonder if it's too broad and just some static defined options are enough. My worry with the mapper is that people might get tricked to believe that in there they can run non deterministic code... Plus if we start from static config, we can evolve it to mapper anyway when moving forward.

PR #542, opened 2025-05-16
Add missing exports

nikrooz:
Adding the missing exports caught by https://github.com/restatedev/sdk-typescript/pull/541

### DIFF
--- a/packages/restate-sdk-testcontainers/package.json
+++ b/packages/restate-sdk-testcontainers/package.json
@@ -46,7 +46,7 @@
     "release": "release-it"
   },
   "dependencies": {
-    "@restatedev/restate-sdk-core": "^1.5.4",
+    "@restatedev/restate-sdk": "^1.5.4",
     "@restatedev/restate-sdk-clients": "^1.5.4",
     "apache-arrow": "^18.0.0",
     "testcontainers": "^10.24.1"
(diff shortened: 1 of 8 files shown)


packages/restate-sdk-testcontainers/package.json
```diff
@@ -46,7 +46,7 @@
     "release": "release-it"
   },
   "dependencies": {
-    "@restatedev/restate-sdk-core": "^1.5.4",
+    "@restatedev/restate-sdk": "^1.5.4",
```
nikrooz, line 49: We were missing `@restatedev/restate-sdk` as a dependency because we use `endpoint` from this package. Since all the types we needed from `@restatedev/restate-sdk-core` are available in `@restatedev/restate-sdk`, I removed the former.

packages/restate-sdk-testcontainers/src/restate_test_environment.ts
```diff
@@ -11,8 +11,15 @@
...
-import * as core from "@restatedev/restate-sdk-core";
+import { endpoint, serde } from "@restatedev/restate-sdk";
+import type {
+  TypedState,
+  UntypedState,
+  Serde,
+  RestateEndpoint,
+  VirtualObjectDefinition,
+  WorkflowDefinition,
+} from "@restatedev/restate-sdk";
```
nikrooz, line 22: Importing with `import * as restate from "@restatedev/restate-sdk"` is vague because it does not specify which parts of `@restatedev/restate-sdk` will be used. Therefore, `api-extractor` requires exporting the entire `restate`, which is why I used named imports here.

Screenshot 2025-05-16 at 09 53 33

PR #550, opened 2025-06-12
Duration data type

slinkydeveloper:


packages/restate-sdk/src/context_impl.ts
```diff
@@ -235,8 +236,8 @@ export class ContextImpl implements ObjectContext, WorkflowContext {
       const parameter = requestSerde.serialize(send.parameter);
 
       let delay;
-      if (send.delay !== undefined) {
-        delay = BigInt(send.delay);
+      if (delay !== undefined) {
+        delay = BigInt(millisOrDurationToMillis(send.delay as Duration));
```
nikrooz, line 240: I think if I set the duration to `1.2345` which there is nothing prevents me from doing so, this will break. probably should make sure we pass an integer to BigInt

nikrooz (approved):
just one comment, other than that, lgtm.

packages/restate-sdk/src/context_impl.ts
slinkydeveloper: Done

PR #561, opened 2025-06-28
Resolve leak on aborted invocation

slinkydeveloper:
On AbortError we were not removing the logger, causing a leak.


slinkydeveloper:
Hey @nikrooz can you just double check this change? I'm also running perf tests now https://github.com/restatedev/e2e-verification-runner/actions/runs/15938903476/job/44963953868

slinkydeveloper:
Performance tests are green, this seems to be the leak!

PR #562, opened 2025-06-30
`asTerminalError` to provide an easy mapping between domain model errors and Restate `TerminalError`

slinkydeveloper:


slinkydeveloper:
I guess you don't want your domain types to depend on restate, or even worse you consume errors from 3rd party libraries. One can always wrap, that's true, but that's not nice either.

nikrooz (approved):
That's a fair. Although `asTerminalError` is kind of a wrapper.

just merely a suggest, instead of `asTerminalError`, we can have `terminateOnError: (error) => boolean`

marcus-sa:
@slinkydeveloper what about this? https://discord.com/channels/1128210118216007792/1388983881050493109

nikrooz:
- For the second part, I was thinking we would pass the user error as the `cause` to our errors (like TerminalError), which can then be sent to the server and journal, including as much context as needed.
- But that still leaves the first question: given a user error, what should the message and errorCode of TerminalError be? The message is easier to handle since the native Error has a message property, and we can generally rely on user errors having one too. If not, we can fall back to a default message. The error code, however, isn’t standardized. I’m starting to think asTerminalError might actually be the better approach here :)

PR #569, opened 2025-07-07
Introduce RetryableError

### LINKED ISSUE #563: Add `RetryAfterError` to support more dynamic retry policies.
See https://discord.com/channels/1128210118216007792/1387936039443894323/1387936039443894323

slinkydeveloper:
Allowing users to customize the retry after when needed.
This is meant to be supported within `ctx.run` primarily.

Fix #563


slinkydeveloper:
Hey @nikrooz can you check this out? I have to be honest, i'm not 100% happy with the name, so if you have proposals feel free to post them.

I didn't go with `RetryAfterError` because i didn't find this nice to read:

```
RetryAfterError.from(cause, { retryAfter: duration })
```

VS

```
RetryableError.from(cause, { retryAfter: duration })
```

nikrooz:
> Except TerminalError itself i suppose?

yes.

> Should those "dynamic" retryAfter and the "static" maxRetryAttempts be merged?

how would you merge it?
```ts
ctx.run(() => {
  /// other stuff
  const res = fetch(...);
  if (!res.ok) {
     throw RetryableError.from(cause, { retryAfter: blabla })
  }
}, {maxRetryAttempts: 2})
```
This makes sense to me. The `ctx.run` will only retry a maximum of two times. If the fetch fails, based on the API response, we may want to wait before retrying.

nikrooz (approved):
lgtm, thanks

PR #576, opened 2025-07-30
Deprecate `endpoint()` API

nikrooz:
We should not merge/publish this, until we updated docs and examples.


nikrooz:
I was holding off on deprecating them until the documentation was updated, but if that’s not required, we can merge.

PR #613, opened 2025-10-27
Use unrepresentable with zod toJSONSchema

slinkydeveloper:


nikrooz (approved):
Lgtm. 

Maybe we can also add 

https://zod.dev/json-schema?id=override

```

override: (ctx) => {
    const def = ctx.zodSchema._zod.def;
    if (def.type === "date") {
      ctx.jsonSchema.type = "string";
      ctx.jsonSchema.format = "date-time";
    }
  },

```

PR #619, opened 2025-11-10
Reorg monorepo

nikrooz:
- use pnpm workspaces
- use turbo for caching tasks
- use plop.js for generating new libs
- use changeset for release

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -62,11 +62,11 @@ For development, you can use watch mode which provides instant type checking wit
 # Watch all lib packages (type checking only)
 pnpm dev
 
-# In another terminal, run examples in dev mode (uses source directly)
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.md
```diff
@@ -1,74 +1,345 @@
...
+### Watch Mode (Recommended)
+
+For development, you can use watch mode which provides instant type checking without builds:
+
+```bash
+# Watch all lib packages (type checking only)
+pnpm dev
+
+# In another terminal, run examples in dev mode (uses source directly)
+pnpm examples:dev
```
slinkydeveloper: This is not really useful, because it's gonna start all examples at the same time, which ofc will crash for port allocation.

Can this command just be a shortcut for the example below.

PR #645, opened 2026-02-18
Add request signing tests to multi-runtime compatibility CI

tillrohrmann:
I've asked our automation engineer Claude to help me with adding support for testing request signing across the different TS runtimes.

Enable identity verification (request signing) in the 5 runtime compatibility tests (Node, Bun, Deno, Cloudflare Workers, Vercel) to ensure that dependency upgrades in restate-sdk-shared-core won't break JWT signature verification across JS runtimes.

Changes:
- Add a test-only ED25519 keypair in test/fixtures/request-signing/
- Modify all 6 example services to optionally enable identity verification via the RESTATE_IDENTITY_KEYS env var
- Switch 5 CI workflow templates from services: block to docker run so the Restate server's signing private key PEM file can be mounted
- Lambda template is unchanged (uses same WASM core as Node)


packages/examples/bun/src/index.ts
```diff
@@ -24,7 +24,10 @@ const greeter = restate.service({
   },
 });
 
+const identityKeys =
+  process.env.RESTATE_IDENTITY_KEYS?.split(",").filter(Boolean);
```
tillrohrmann: Is this ok to add to the public examples? Alternatively, we could add dedicated testing handlers to the repo. Same question for all the other example handlers.

slinkydeveloper: sounds good to me, these are not ready to clone examples anyway, that's something we have in the examples repo.
just make sure it tolerates starting without identity keys.

nikrooz (approved):
LGTM

PR #650, opened 2026-02-21
Expose cancellation signal on request

mcuelenaere:
## Summary

This PR adds a `cancellationSignal: AbortSignal` property to the `Request` interface that aborts as soon as the Restate runtime sends a cancellation, rather than waiting for the next `ctx.run()` call.

This allows passing the signal to `fetch()`, database clients, or any API that accepts an `AbortSignal` for proactive cancellation of in-flight operations. The signal's reason is a `CancelledError`.

## Motivation

When an invocation is cancelled, the SDK currently only detects this at the next Restate operation. For handlers with long-running user code between operations (e.g., AI/LLM streaming), cancellation detection can be significantly delayed.

## Implementation

A new `CancellationWatcherPromise` hooks into the existing `PromisesExecutor` loop to monitor `is_completed(cancel_handle())` and abort an `AbortController` when cancellation is detected. The watcher is stopped via `invocationEndPromise` when the invocation ends, preventing interaction with a closed VM.

[Design document](https://gist.github.com/mcuelenaere/e1bb7390720aa3b4e0aa0c5973e71314)

## Notes

I've verified this change works correctly in my local setup. The added E2E tests I haven't run yet, I was hoping the CI could help with that.

### DIFF
--- a/packages/libs/restate-sdk/src/context.ts
+++ b/packages/libs/restate-sdk/src/context.ts
@@ -76,6 +76,15 @@ export interface Request {
    * than cleanup any external resources that might be shared across attempts (e.g. database connections).
    */
   readonly attemptCompletedSignal: AbortSignal;
+
+  /**
+   * Signal that is aborted when the invocation is cancelled by the Restate runtime.
+   * Unlike {@link attemptCompletedSignal}, this signal specifically indicates cancellation
+   * and can be used to abort in-flight operations (e.g., fetch calls, database queries).
+   *
+   * The signal's reason will be a {@link CancelledError}.
+   */
+  readonly cancellationSignal: AbortSignal;
 }
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
--- a/packages/libs/restate-sdk/src/context_impl.ts
+++ b/packages/libs/restate-sdk/src/context_impl.ts
@@ -34,6 +34,7 @@ import {
   WasmHeader,
 } from "./endpoint/handlers/vm/sdk_shared_core_wasm_bindings.js";
 import {
+  CancelledError,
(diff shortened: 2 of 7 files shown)


tillrohrmann:
Thanks a lot for your contribution @mcuelenaere. I think this is a really nice improvement. Curious to hear what @nikrooz thinks about this idea.

The one thing that one needs to be aware of is that with this change, the code snippet that reacts to the `AbortSignal` might have done some side effects and we record the journal entry as completed with a `CancellationException`. Hence, `CancellationException` will no longer mean that this journal entry hasn't been executed. This is not different from the `ctx.run` failing with a terminal exception. Another aspect is that we will now have potentially a cancelled journal entry as well as the cancellation signal in the journal. @slinkydeveloper is this a problem?

mcuelenaere:
hi @tillrohrmann, I didn't think of any possible side effects that could occur from this. To give some context, our usecase is a restate service that does streaming LLM (via an out-of-band pub-sub mechanism), but we'd like the user to be able to abort this. My thinking was to leverage the built-in Restate cancellation mechanism, but we do need to do some simple cleanup/bookkeeping (setting some state in the virtual object) after cancellation is requested. If that breaks certain Restate assumptions, then I could look into pivotting the cancellation to an out-of-band mechanism as well (and drop this PR). WDYT?

slinkydeveloper:
I'm going to close this, as we're now going to expose in `ContextInternal` a new functionality that allows you to get the "cancellation" `RestatePromise` directly, and inside `.map` you can easily hook up an abort controller. This should reach the goal of this PR, while being safely deterministic. Thanks for starting the discussion on the topic @mcuelenaere

PR #685, opened 2026-04-13
[WIP] `.map` now can be async

slinkydeveloper:


slinkydeveloper:
now that we have gen sdk this is not relevant anymore!

Issue #690, opened 2026-04-16
Bug: [restate-sdk-clients] `workflowSubmit` ignores SendOpts when workflow has no input parameter


Nez21:
When a workflow's run handler takes no input, calling `workflowSubmit(sendOpts)` passes `SendOpts` as the parameter arg — not as opts. The send options are serialized as the request body rather than applied as invocation options (delay, headers, idempotency key, etc.).

```typescript
// Workflow with no input
const noInputWorkflow = restate.workflow({
    name: 'no-input-workflow',
    handlers: {
        run: async (ctx) => {},
    },
})

client.workflowClient(noInputWorkflow, 'workflow-key').workflowSubmit(rpc.sendOpts({ delay: 5000 }));
// BUG: delay is ignored, SendOpts object sent as request body
```

PR #691, opened 2026-04-16
fix: update workflowSubmit to use optsFromArgs so it can handle opts …

Nez21:
PR to fix issue: #690


nikrooz (approved):
@Nez21 Thanks for fixing this, and sorry it took so long to look at it.

The changes look good to me, will merge it.

PR #695, opened 2026-04-22
Move response-head commit inside process()

nikrooz:
## Problem
  `RestateResponse` exposed `statusCode` / `headers` as eager fields fixed at
  `handle()` time. That prevented adding routes whose response head depends on
  parsing the request body.

 ## Change
  `process()` now receives a `writeHead(statusCode, headers)` callback and
  commits the head at a point of its choosing. A safety layer in
  `RestateHandler.handle()` guarantees `writeHead` is called exactly once,
  emits a 500 fallback on pre-commit failures, and closes the output stream
  on every path. Adapters become pure transport glue; `/invoke`'s streaming
  flow is unchanged.

### DIFF
--- a/packages/libs/restate-sdk/src/endpoint/handlers/utils.ts
+++ b/packages/libs/restate-sdk/src/endpoint/handlers/utils.ts
@@ -12,6 +12,7 @@ import type {
 import { createLogger, Logger } from "../../logging/logger.js";
 import { parseUrlComponents } from "../components.js";
 import { X_RESTATE_SERVER } from "../../user_agent.js";
+import { CompletablePromise } from "../../utils/completable_promise.js";
 
 export function tryCreateContextualLogger(
   loggerTransport: LoggerTransport,
@@ -71,15 +72,13 @@ export function simpleResponse(
   body: Uint8Array
 ): RestateResponse {
   return {
-    headers,
-    statusCode,
-    async process({ inputReader, outputWriter }): Promise<void> {
-      if (inputReader !== undefined) {
-        // Drain the input stream
-        while (true) {
-          const { done } = await inputReader.next();
-          if (done) break;
-        }
+    async process({ inputReader, outputWriter, writeHead }): Promise<void> {
+      writeHead(statusCode, headers);
+
+      // Drain the input stream
+      while (true) {
+        const { done } = await inputReader.next();
+        if (done) break;
       }
 
       await outputWriter.write(body);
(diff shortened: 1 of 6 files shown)


packages/libs/restate-sdk/src/endpoint/handlers/utils.ts
```diff
@@ -92,3 +91,27 @@ export function simpleResponse(
...
+ * commit from outside the `process()` call.
+ *
+ * `writeHead` is expected to be called exactly once — enforced by the
+ * safety layer in {@link RestateHandler.handle}, so there is no guard here.
+ */
+export function captureHead(): {
+  writeHead: (statusCode: number, headers: ResponseHeaders) => void;
+  head: Promise<{ statusCode: number; headers: ResponseHeaders }>;
+} {
+  const ready = new CompletablePromise<{
```
slinkydeveloper: absolute side note, i wonder when we can get rid of this and use Promise.withResolvers()

slinkydeveloper (approved):
as discussed offline, let's remove the safety net if we don't need it. except that, lgtm

packages/libs/restate-sdk/src/endpoint/handlers/utils.ts
nikrooz: done

nikrooz:
>as discussed offline, let's remove the safety net if we don't need it. except that, lgtm

@slinkydeveloper done

packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
```diff
@@ -176,29 +180,37 @@ function nodeHandlerImpl(
...
 
     // Abort controller used to cleanup resources at the end of this stream lifecycle
     const abortController = new AbortController();
     httpRequest.on("close", () => {
       abortController.abort();
     });
 
+    const writeHead = (statusCode: number, headers: ResponseHeaders): void => {
+      res.writeHead(statusCode, headers);
+    };
```
slinkydeveloper: js noob question here: isn't it better here to do `res.writeHead.bind(res)` to avoid capturing less stuff?

slinkydeveloper (approved):
just a lil question, otherwise let's goooo

packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
nikrooz: yes, much cleaner

PR #701, opened 2026-04-28
Add testcontainer networking and storage options

nikrooz:
Add testcontainer networking and storage options

Remove the deprecated mount callback start API and fold the container, endpoint access, and storage settings into TestEnvironmentOptions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ lambda.zip
 
 .turbo
 .npm-cache
+napkin.md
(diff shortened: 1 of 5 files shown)


igalshilman (approved):
Looks good!
But I’m thinking maybe we actually switch the behavior for testing to the faster option?
Otherwise no one would know about it?

nikrooz:
@igalshilman  I agree we should make that the default, but I assumed it wouldn’t work for all cases; otherwise, Testcontainers would already be using it.

If that’s correct, then this would be a breaking change, so I thought we could first introduce the option and then change the default later.

But if you think that’s too conservative, I’m happy to change the default now.

PR #707, opened 2026-05-01
Memory leak tests + fix

nikrooz:
Adds a test to capture memory leak in both node and fetch handler
1. record the heap size
2. make 125 invocations (configurable) with various statuses (failed, succeed, retrying,...)
3. kill all
4. measure heap and calculate the delta
repeat for 20 times (configurable)

a sample PASS result
```
 PASS SDK memory delta check
 observed heap delta: 0.9 MiB
 threshold: 8.0 MiB
 invocations: 2500

 Memory by iteration:
 | iteration | invocations | heap after GC | delta (vs prev) | delta (vs baseline) | heap before cleanup |
 | --------- | ----------- | ------------- | --------------- | ------------------- | ------------------- |
 | baseline  | 0           | 10.9 MiB      |                 | 0.0 MiB             |                     |
 | 1         | 125         | 11.1 MiB      | 0.2 MiB         | 0.2 MiB             | 11.4 MiB            |
 | 2         | 125         | 11.2 MiB      | 0.0 MiB         | 0.2 MiB             | 11.5 MiB            |
 | 3         | 125         | 11.4 MiB      | 0.2 MiB         | 0.4 MiB             | 12.0 MiB            |
 | 4         | 125         | 11.4 MiB      | 0.1 MiB         | 0.5 MiB             | 12.1 MiB            |
 | 5         | 125         | 11.5 MiB      | 0.0 MiB         | 0.5 MiB             | 12.1 MiB            |
 | 6         | 125         | 11.6 MiB      | 0.1 MiB         | 0.6 MiB             | 12.1 MiB            |
 | 7         | 125         | 11.6 MiB      | 0.0 MiB         | 0.7 MiB             | 12.1 MiB            |
 | 8         | 125         | 11.6 MiB      | -0.0 MiB        | 0.6 MiB             | 12.2 MiB            |
 | 9         | 125         | 11.5 MiB      | -0.0 MiB        | 0.6 MiB             | 12.1 MiB            |
 | 10        | 125         | 11.6 MiB      | 0.1 MiB         | 0.7 MiB             | 12.2 MiB            |
 | 11        | 125         | 11.7 MiB      | 0.1 MiB         | 0.7 MiB             | 12.2 MiB            |
 | 12        | 125         | 11.8 MiB      | 0.1 MiB         | 0.8 MiB             | 12.4 MiB            |
 | 13        | 125         | 11.8 MiB      | 0.0 MiB         | 0.8 MiB             | 12.3 MiB            |
 | 14        | 125         | 11.7 MiB      | -0.1 MiB        | 0.8 MiB             | 12.4 MiB            |
 | 15        | 125         | 11.7 MiB      | 0.0 MiB         | 0.8 MiB             | 12.3 MiB            |
 | 16        | 125         | 11.8 MiB      | 0.1 MiB         | 0.9 MiB             | 12.4 MiB            |
 | 17        | 125         | 11.8 MiB      | -0.0 MiB        | 0.8 MiB             | 12.4 MiB            |
 | 18        | 125         | 11.7 MiB      | -0.0 MiB        | 0.8 MiB             | 12.4 MiB            |
 | 19        | 125         | 11.8 MiB      | 0.0 MiB         | 0.9 MiB             | 12.4 MiB            |
 | 20        | 125         | 11.8 MiB      | 0.0 MiB         | 0.9 MiB             | 12.5 MiB            |
 
 Final heap summary:
 | metric                   | value    |
 | ------------------------ | -------- |
 | invocations              | 2500     |
 | baseline heap            | 10.9 MiB |
 | final heap after GC      | 11.8 MiB |
 | heap delta (vs baseline) | 0.9 MiB  |
 | retained heap threshold  | 8.0 MiB  |
 ```

A sample FAILED result
```
 FAIL SDK memory delta check
 observed heap delta: 27.4 MiB
 threshold: 8.0 MiB
 invocations: 2500
 
 Memory by iteration:
 | iteration | invocations | heap after GC | delta (vs prev) | delta (vs baseline) | heap before cleanup |
 | --------- | ----------- | ------------- | --------------- | ------------------- | ------------------- |
 | baseline  | 0           | 12.4 MiB      |                 | 0.0 MiB             |                     |
 | 1         | 125         | 14.2 MiB      | 1.8 MiB         | 1.8 MiB             | 43.1 MiB            |
 | 2         | 125         | 15.2 MiB      | 1.1 MiB         | 2.8 MiB             | 37.3 MiB            |
 | 3         | 125         | 16.7 MiB      | 1.5 MiB         | 4.3 MiB             | 39.4 MiB            |
 | 4         | 125         | 18.2 MiB      | 1.5 MiB         | 5.8 MiB             | 39.4 MiB            |
 | 5         | 125         | 19.4 MiB      | 1.2 MiB         | 7.0 MiB             | 39.5 MiB            |
 | 6         | 125         | 21.2 MiB      | 1.8 MiB         | 8.8 MiB             | 47.9 MiB            |
 | 7         | 125         | 22.5 MiB      | 1.3 MiB         | 10.1 MiB            | 30.9 MiB            |
 | 8         | 125         | 23.8 MiB      | 1.4 MiB         | 11.4 MiB            | 24.5 MiB            |
 | 9         | 125         | 25.3 MiB      | 1.5 MiB         | 12.9 MiB            | 49.5 MiB            |
 | 10        | 125         | 26.7 MiB      | 1.4 MiB         | 14.4 MiB            | 32.0 MiB            |
 | 11        | 125         | 28.0 MiB      | 1.3 MiB         | 15.6 MiB            | 49.8 MiB            |
 | 12        | 125         | 29.3 MiB      | 1.3 MiB         | 16.9 MiB            | 50.6 MiB            |
 | 13        | 125         | 30.5 MiB      | 1.2 MiB         | 18.1 MiB            | 60.1 MiB            |
 | 14        | 125         | 31.9 MiB      | 1.5 MiB         | 19.5 MiB            | 56.1 MiB            |
 | 15        | 125         | 33.2 MiB      | 1.3 MiB         | 20.8 MiB            | 43.0 MiB            |
 | 16        | 125         | 34.4 MiB      | 1.2 MiB         | 22.0 MiB            | 55.9 MiB            |
 | 17        | 125         | 35.6 MiB      | 1.3 MiB         | 23.2 MiB            | 57.6 MiB            |
 | 18        | 125         | 37.0 MiB      | 1.4 MiB         | 24.6 MiB            | 57.9 MiB            |
 | 19        | 125         | 38.5 MiB      | 1.4 MiB         | 26.1 MiB            | 58.4 MiB            |
 | 20        | 125         | 39.8 MiB      | 1.3 MiB         | 27.4 MiB            | 52.8 MiB            |
 
 Final heap summary:
 | metric                   | value    |
 | ------------------------ | -------- |
 | invocations              | 2500     |
 | baseline heap            | 12.4 MiB |
 | final heap after GC      | 39.8 MiB |
 | heap delta (vs baseline) | 27.4 MiB |
 | retained heap threshold  | 8.0 MiB  |
 
```

**Problem:**
As raised in #706, Fetch invocations could retain per-invocation SDK state after completion. The fetch request signal could keep invocation abort listeners alive, and the SDK’s module-level invocation logger map was not cleared on normal completion.

**Fix:**
Use a local AbortController in the fetch adapter, forward request aborts into it, and abort it when processing completes. Also remove the invocation logger after flushAndClose() on all completion paths.

### CODE REVIEW

File: packages/libs/restate-sdk/src/endpoint/node_endpoint.ts [hunk 1]
```diff
@@ -213,6 +212,10 @@ function nodeHandlerImpl(
             httpRequest.headers
           ) ?? endpoint.rlog;
         logger.error("Unexpected error: " + (error.stack ?? error.message));
+      })
+      .finally(() => {
+        httpRequest.off("close", abort);
```
slinkydeveloper: do you need to trigger this yourself? could that cause issues?

packages/libs/restate-sdk/src/endpoint/handlers/fetch.ts
```diff
@@ -39,13 +39,21 @@ export function fetcher(handler: RestateHandler) {
...
+      } else {
+        event.signal.addEventListener("abort", abort, { once: true });
+      }
 
       response
         .process({
           inputReader,
           outputWriter,
           writeHead,
-          abortSignal: event.signal,
```
slinkydeveloper, line 48: IMO this is a bun bug. My expectation is that the abort signal is fired neverthless, but clearly it isn't.

slinkydeveloper (commented):
a bit puzzled by this....

.github/workflows/integration.yaml
slinkydeveloper, line 1: instead of updating this file, can you create another integration.yaml file for fetch/bun? This file is used by runtime too, i don't think we need to run bun tests in runtime. Also two files -> easier to mantain

packages/libs/restate-sdk/src/endpoint/handlers/generic.ts
```diff
@@ -469,12 +469,16 @@ class RestateInvokeResponse implements RestateResponse {
...
-      );
+      try {
+        await flushAndClose(
+          this.coreVm,
+          this.vmLogger,
+          inputReader,
+          outputWriter
+        );
+      } finally {
+        destroyLogger(this.loggerId);
```
slinkydeveloper, line 483: I understand you're adding this as safety net, but the contract of this RestateResponse is that the abort controller is called neverthless. I think i even wrote it in the docs.

packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
slinkydeveloper: Given we know it works on node, can we leave the code as it was before? or is there any reason for this addition?

packages/libs/restate-sdk/src/endpoint/handlers/fetch.ts
nikrooz: I might be wrong, but I don’t think this is a Bun bug.

- I tested a few fetch-style runtimes/adapters. Bun, Hono’s Node adapter, Cloudflare Workers/Miniflare, Vercel/Next Node route handlers, and Vercel/Next Edge route handlers do not abort `Request.signal` after a successful response. Deno does, but that seems to be Deno-specific behavior, not something we can rely on everywhere.
- My understanding is that `Request.signal` means “the request was cancelled/aborted”, not “the handler finished successfully”. Some runtimes may also abort it when the request finishes, but that is not portable behavior.
- So if we pass `abortSignal: event.signal`, then `attemptCompletedSignal` behaves differently depending on the runtime.
- I think the fetch adapter should create its own `AbortController`, forward `event.signal` aborts into it, and abort it itself when `response.process()` finishes.

.github/workflows/integration.yaml
nikrooz: sure, will do that.

packages/libs/restate-sdk/src/endpoint/handlers/generic.ts
nikrooz: yes, it's just a safety net. I thought it would make things simpler if the cleanup does not rely on external signal. but happy to revert it you prefer

packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
nikrooz: no we don't need it, I'll revert it. it might be good to keep `httpRequest.off("close", abort);`.
for now I revert it completely, but if you are happy with removing listener I'll add it back.

.github/workflows/integration-fetch.yaml
```diff
@@ -0,0 +1,204 @@
...
+        description: "restate image, superseded by restate commit"
+        required: false
+        default: "ghcr.io/restatedev/restate:main"
+        type: string
+      serviceImage:
+        description: "service image, if provided it will skip building the image from sdk main branch"
+        required: false
+        default: ""
+        type: string
+  workflow_call:
```
slinkydeveloper: we don't need call on this one right? just dispatch (the one i can trigger from ui)

packages/libs/restate-sdk/src/endpoint/handlers/generic.ts
slinkydeveloper: look i'm ok if you wanna keep it, but document it. otherwise when we'll go back at this code we'll be wondering what's going on here with calling destroyLogger twice :D

packages/libs/restate-sdk/src/endpoint/handlers/fetch.ts
```diff
@@ -39,13 +39,21 @@ export function fetcher(handler: RestateHandler) {
...
 
       const { writeHead, head } = captureHead();
+      // propagate request aborts, and abort again when processing completes.
+      const abortController = new AbortController();
+      const abort = () => abortController.abort();
+      if (event.signal.aborted) {
+        abort();
+      } else {
+        event.signal.addEventListener("abort", abort, { once: true });
+      }
```
slinkydeveloper: can you add some comment here about the stuff you mentioned that different runtimes do whatever they want here with this signal?

slinkydeveloper (approved):
left few bits, otherwise lgtm

PR #718, opened 2026-05-13
Scope and limit key API

slinkydeveloper:


packages/libs/restate-sdk/src/context.ts
```diff
@@ -251,12 +268,36 @@ export type GenericSend<REQ> = {
   inputSerde?: Serde<REQ>;
   delay?: Duration | number;
   idempotencyKey?: string;
+  scope?: string;
```
igalshilman, line 298: I think we need to put here a tsdoc comment and an attribute 'experimental` and say that it is only applicable starting with 1.7 with vqueues enabled.
alternatively we can follow aisdk convention and name these `experimental_scope` and `experimental_limitkey` (not a fan of this)

igalshilman (approved):

packages/libs/restate-sdk/src/context.ts
slinkydeveloper: not a fan of calling this `experimental` either, for me we can ship it like this adding tsdoc comments.

slinkydeveloper: updated

PR #736, opened 2026-06-05
[restate-sdk-gen] Abandon spawned fibers when the main fiber exits (onMainExit)

igalshilman:
## What

Changes the default lifetime of spawned fibers in `@restatedev/restate-sdk-gen`: when the **main fiber exits, all other fibers are stopped** ("abandoned"). Previously `Scheduler.run()` only returned once *every* fiber had finished, so a spawned fiber (or race loser) parked on a never-settling source kept the handler alive forever — the documented race-loser footgun.

The behavior is configurable via a new option on `execute`:

```ts
execute(ctx, op);                          // default: { onMainExit: "abandon" }
execute(ctx, op, { onMainExit: "join" });  // old wait-for-all behavior
```

## Semantics of `"abandon"` (the new default)

- `run()` returns as soon as the main fiber settles; the stop is **prompt** — `drainReady()` halts mid-queue, so nothing observable (journal writes, channel sends, side effects) happens after the main fiber's outcome is decided.
- Abandoned fibers are never resumed: no `catch`/`finally` runs, and the sources they were parked on are dropped. Durable work already performed stays journaled; only the in-memory continuation is discarded.
- The stop-point is **replay-deterministic**: per-tick race winners are journaled combinator entries, so replay walks the same trajectory and stops at the same point (pinned by a new `alwaysReplay` e2e).
- ⚠️ **BREAKING (behavior)**: fire-and-forget spawns are no longer guaranteed to complete. `yield*` the spawned future before returning, or pass `{ onMainExit: "join" }`.

## API surface

- New `OnMainExit = "abandon" | "join"` + `SchedulerOptions` (scheduler.ts), `ExecuteOptions` alias (restate-operations.ts).
- `execute` now takes an optional third options argument and is **exported from the package index** (it was always documented as public in `free.ts` but never actually exported), together with the new option types.

## Testing

- `test/on-main-exit.test.ts`: abandon/join semantics, prompt mid-drain stop, combinator-fallback abandonment, cancellation interplay, main-rejection path; `spawn`/`mixing-select` tests that pinned wait-for-all now opt into `"join"` explicitly. 251/251 unit tests pass.
- `e2e/abandon.e2e.test.ts`: fire-and-forget multi-step spawn and stuck race loser against a real runtime — 4/4 pass locally in both `default` and `alwaysReplay` modes (suspend → full replay → success, validating replay determinism of the abandon boundary).
- `pnpm verify` green across the workspace (64/64 tasks; deno example excluded locally — no deno binary on the dev machine).
- Adversarially reviewed via parallel reviewer agents (replay determinism, cancellation/scheduler edge cases, API/downstream impact): all e2e services, test-services (incl. the conformance `vo-command-interpreter`), bench, and tutorial either await their spawns transitively or only read race winners, so no existing behavior breaks.

## Docs

README.md, DESIGN.md (new *"When the main operation settles: `onMainExit`"* section with design rationale), guide.md (*"Race losers and spawn lifetime"*), tutorial comments (`02-spawn.ts`, `06-cancel.ts`), and a changeset (currently `minor` with a BREAKING note — happy to flip to `major` if preferred).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/libs/restate-sdk-gen/src/scheduler.ts
+++ b/packages/libs/restate-sdk-gen/src/scheduler.ts
@@ -87,15 +87,18 @@ export class Scheduler implements SchedulerOps {
   // still see it as aborted (correctly, since their work was caught
   // by the cancellation); closures created after recovery see the new,
   // unaborted signal.
-  private abortController: AbortController = new AbortController();
+  private abortController: AbortController;
+
   /**
-   * External AbortSignals linked via {@link linkAbortSignal}. When any
-   * of them fires, the *current* controller is aborted; replacement
-   * controllers (cancellation recovery) are born pre-aborted if a
-   * linked signal has already fired — an attempt that is over stays
-   * over, unlike cancellation, which is recoverable.
+   * Parent AbortSignal supplied at construction (production: the SDK's
+   * `attemptCompletedSignal`). Every controller this scheduler creates
+   * is a child of it — born aborted if the parent has already fired,
+   * aborted the moment it fires otherwise. Unlike cancellation, a
+   * parent abort is sticky: an ended attempt never resumes. Defaults
+   * to a signal that never fires.
    */
-  private readonly linkedSignals: AbortSignal[] = [];
+  private readonly parentSignal: AbortSignal;
+
   /**
    * Slot for the operations object bound to this scheduler. Defaults
    * to the scheduler itself so tests (which don't construct a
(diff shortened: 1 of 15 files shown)


packages/libs/restate-sdk-gen/src/scheduler.ts
```diff
@@ -43,10 +43,42 @@ import type { Settled, PromiseSource } from "./scheduler-types.js";
...
+ * - `"join"`: `run()` keeps driving until *every* routine has
+ *   finished. A spawned routine parked on a source that never settles
+ *   keeps the handler from returning forever — prefer `"abandon"`
+ *   unless the workflow relies on fire-and-forget routines completing.
+ */
+export type OnMainExit = "abandon" | "join";
+
+export interface SchedulerOptions {
+  /** See {@link OnMainExit}. Defaults to `"abandon"`. */
+  onMainExit?: OnMainExit;
```
slinkydeveloper, line 67: i think `main` is confusing.

how about `onInvocationEnd`?

PR #749, opened 2026-06-15
[restate-sdk-tunnel] Client-initiated graceful shutdown

igalshilman:
## What

Lets an in-process tunnel deployment **drain gracefully on `SIGTERM`** — stop accepting new invocations, finish in-flight ones, then exit — so a rolling restart drops no invocations. This is the client half; the server half is restatedev/restate-cloud#958.

## How

`SendRequest::is_ready()` on the server doesn't reflect an h2 GOAWAY until the connection fully closes, so a draining client can't deselect itself at the h2 layer. Instead the client signals drain at the application level:

- Advertises **`supports-client-drain: true`** in the handshake (new `supportsClientDrain` option, default `true`).
- On `shutdown()`: stops the supervisor (no new dials), then refuses each **new** forwarded invocation with **`503` + `x-restate-tunnel-draining: true`** *without running the handler*. The server deselects this connection and retries that invocation on a healthy replica — no double-execution, since the handler never ran.
- **In-flight** invocations keep running and are awaited (bounded by `graceMs`, default `drainGraceMs` = 120s); then the connection is torn down. An **idle** connection needs no signal — it just closes and the server's `TunnelGuard` removes it.

## API

- `TunnelConnection.shutdown({ graceMs? }): Promise<void>` — drain, then stop. Wire it to `SIGTERM` (`process.on("SIGTERM", () => conn.shutdown().then(() => process.exit(0)))`). Degrades to an abrupt `close()` if `supportsClientDrain` wasn't advertised.
- `gracefulShutdown?: boolean | { signals?, graceMs? }` (off by default) — opt in to auto-installing one-shot signal handlers (default `SIGTERM`) that call `shutdown()` then `process.exit(0)`. A library shouldn't commandeer signals unless asked.
- `supportsClientDrain?: boolean` (default `true`) — advertise the capability.

No backward-compat constraints (package has no released users yet).

## Testing

`tsc --noEmit`, `eslint .`, and `vitest run` (92/92, incl. new `test/shutdown.test.ts`: capability advertisement, idle fast-path / no-redial, and draining in-flight while refusing new invocations with the sentinel) all pass on Node v24.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/libs/restate-sdk-tunnel/src/connect.ts
+++ b/packages/libs/restate-sdk-tunnel/src/connect.ts
@@ -119,12 +119,14 @@ export function connectTunnel(options: ConnectTunnelOptions): TunnelConnection {
   const draining = new DrainingRegistry();
   const inflight = new InflightTracker();
 
-  // `shuttingDown` makes every connection refuse new invocations with the
-  // drain sentinel (so the cloud deselects us). `stopped`/`toreDown` guard the
-  // handle methods against re-entry.
-  let stopped = false;
-  let toreDown = false;
-  let shuttingDown = false;
+  // The engine lifecycle as one disjoint state rather than a set of booleans
+  // whose combinations could express nonsense:
+  //   running  — normal operation.
+  //   draining — shutdown() in progress: every connection refuses new
+  //              invocations with the drain sentinel (so the cloud deselects
+  //              us) while in-flight ones finish.
+  //   closed   — torn down (abrupt close(), or a completed / grace-expired drain).
+  let phase: "running" | "draining" | "closed" = "running";
   let connectionCount = 0;
   let lastInfo: HandshakeInfo | undefined;
 
@@ -150,7 +152,7 @@ export function connectTunnel(options: ConnectTunnelOptions): TunnelConnection {
       lastInfo = info;
       ready.resolve();
     },
-    isShuttingDown: () => shuttingDown,
+    isShuttingDown: () => phase === "draining",
     inflightStarted: () => inflight.started(),
     inflightEnded: () => inflight.ended(),
   };
@@ -177,8 +179,8 @@ export function connectTunnel(options: ConnectTunnelOptions): TunnelConnection {
 
   // Abrupt teardown, shared by close() and a grace-expired shutdown(). Idempotent.
   const teardown = (): void => {
-    if (toreDown) return;
-    toreDown = true;
+    if (phase === "closed") return;
+    phase = "closed";
     supervisor.abortAll();
     for (const socket of activeSockets) socket.destroy();
     activeSockets.clear();
@@ -187,7 +189,6 @@ export function connectTunnel(options: ConnectTunnelOptions): TunnelConnection {
   };
 
   const close = async (): Promise<void> => {
-    stopped = true;
     teardown();
     await done;
   };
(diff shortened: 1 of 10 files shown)


igalshilman:
### Follow-up commit: explicit connection state machine (`b09b24d`)

Refactor only — no behavior change (tsc, eslint, 92/92 vitest still green). Reviewer context:

The first commit left `ConnectionAttempt` carrying its lifecycle in a set of booleans/optionals (`settled`, `serving`, `openedAt`, `handshakePromise`, `detachForDrain`, `session`) that had to be kept mutually consistent by hand — and notably encoded the two drain modes asymmetrically (server drain via `detachForDrain` + a `serving` flag left `true` after `settle`; client drain via an engine-level flag).

This commit replaces that with one explicit `AttemptState` discriminated union:

```
connecting → handshaking{session, handshake?} → serving{session, openedAt}
           → draining{session, openedAt, trigger: "server" | "client"} → closed
```

Key points:

- **Both drains are now the same explicit `draining` state**, distinguished by `trigger`. The *essential* difference is preserved and documented, not hidden: a `server` drain detaches the session to the `DrainingRegistry` and redials (the zero-drop property — the detached session keeps serving its in-flight invocations); a `client` drain refuses new invocations with the sentinel and finishes in place, no redial.
- `run()`-resolution is decoupled from teardown (`resolve()` once vs `settle()` → destroy → `closed`). The server-drain handover resolves `run()` early (so the slot redials) while the registry destroys the session later — previously expressed implicitly by leaving `serving = true` after `settle`.
- Illegal states (e.g. "serving with no session") are now unrepresentable; refusal is gated precisely on `draining{client}` / engine-shutdown, so `draining{server}` still serves.
- The engine gains an `activeConnections` registry so `shutdown()` drives each live connection's `beginClientDrain()`.

Recommend reviewing `connection.ts` as a near-rewrite rather than line-by-line; the diff is large but mechanical.

pcholakov (commented):
I didn't look at the SDK change closely but Codex + Claude reviews came back with these.

---

Consolidated SDK PR #749 findings:

1. **`shutdown()` ignores `supportsClientDrain: false`**
   `packages/libs/restate-sdk-tunnel/src/connect.ts:253`, `src/connection.ts:620`, `src/types.ts:270`
   This is Claude’s finding and it’s legit. The docs promise `shutdown()` degrades to abrupt `close()` when `supportsClientDrain` was not advertised, but the implementation still enters draining and emits `503 + x-restate-tunnel-draining`. Cloud PR #958 correctly ignores that sentinel for non-capable clients, so requests can keep getting routed to the same connection as ordinary 503s. Fix with `if (!opts.supportsClientDrain) return close();` in `shutdown()`, plus a regression test.

2. **`gracefulShutdown` signal handlers are not removed on teardown**
   `packages/libs/restate-sdk-tunnel/src/connect.ts:276`
   The PR installs `process.once(signal, ...)` handlers but never unregisters them on `close()`/teardown. A closed tunnel connection can later intercept `SIGTERM` and call `process.exit(0)`, which is surprising for long-lived processes or tests that create and close connections. Store the installed handlers and remove them during `teardown()`.

packages/libs/restate-sdk-tunnel/src/connect.ts
```diff
@@ -80,194 +144,136 @@ export function connectTunnel(options: ConnectTunnelOptions): TunnelConnection {
...
+    // await, so every connection sees isShuttingDown() for the whole drain.
+    // (drainGracefully runs synchronously up to inflight.whenDrained, so no new
+    // invocation can interleave before `state` is set.)
+    const { active } = state;
+    const completed = drainGracefully(active, graceMs ?? opts.drainGraceMs);
+    state = { kind: "draining", active, completed };
+    return completed;
   };
 
   if (options.signal?.aborted) {
```
pcholakov: Codex called this out:

> the options.signal?.aborted path can still leave a gracefulShutdown process handler installed after teardown. void close() runs teardown() synchronously while signalHandlers is still empty, then the code continues and installs the process.once(...) handlers at lines 288-299. So connectTunnel({ signal: alreadyAbortedSignal, gracefulShutdown: true, ... }) returns a closed connection that can later intercept SIGTERM and call process.exit(0). Move the gracefulShutdown handler registration before the already-aborted signal check, or skip registration when state.kind === "closed" / options.signal.aborted.

pcholakov (approved):
Minor finding from Codex, address if you like otherwise 🚢! 

> Non-blocking: the already-aborted options.signal path can still install gracefulShutdown handlers after close() has already torn down. Consider skipping handler registration when options.signal?.aborted, or registering handlers before the aborted-signal check so teardown removes them.

PR #765, opened 2026-07-12
Allow passing custom fetch implementation to client

### LINKED ISSUE #764: Allow specifying custom fetch implementation in client
Currently, the restate client sdk utilizes `globalThis.fetch` with no option for the user to configure the fetch handler.

This prevents users from leveraging other http clients other than fetch. This may be problematic in cases where additional functionality is needed that native `fetch` cannot provide, like supporting the HTTP/2 Cleartext (H2C) protocol.

n-siddarth:
Fixes #764

### DIFF
--- a/packages/libs/restate-sdk-clients/src/api.ts
+++ b/packages/libs/restate-sdk-clients/src/api.ts
@@ -564,4 +564,11 @@ export type ConnectionOpts = {
    * @experimental
    */
   journalValueCodec?: JournalValueCodec;
+
+  /**
+   * Custom fetch handler
+   *
+   * Defaults to globalThis.fetch
+   */
+  fetch?: typeof globalThis.fetch;
 };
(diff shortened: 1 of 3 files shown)


packages/libs/restate-sdk-clients/src/api.ts
```diff
@@ -564,4 +564,11 @@ export type ConnectionOpts = {
    * @experimental
    */
   journalValueCodec?: JournalValueCodec;
+
+  /**
+   * Custom fetch handler
```
slinkydeveloper:
```suggestion
   * Custom fetch client. This allows you to provide a different fetch implementation.
```

slinkydeveloper (commented):
This is great, thank you. can you just improve the tsdoc a bit to explain why you would provide a different fetch client?

n-siddarth:
Sure, will do that later today 👍

n-siddarth:
@slinkydeveloper Is this better?

slinkydeveloper:
Thanks for the contribution!

PR #766, opened 2026-07-15
Fix premature Node HTTP/1.1 aborts after request body completion

aPoCoMiLogin:
was testing some use-cases with flow-control in nodejs, created simple handler that only had `await setTimeout(5000)` and invoked few tasks, that kept failing with `RestateError: (500) Connection closed` error. was poking around, and found that the abort controller was triggered right after the connection closed, which simply aborted the connection when the request finished. it wasn't visible in http2 or when used with durable wrappers (eg `ctx.run()`).

because this only affects http1.1 i've prepared separate e2e/integration tests, similarly how tests for `fetch` are done. i'm not sure about the github actions, so that would need more attention.

EDIT:

some more information:  `close` event is always triggered by `IncomingMessage` no matter if it was aborted or finished, since node 16: https://nodejs.org/api/http.html#event-close-3 to resolve that issue we can use `readableEnded` of the `IncomingMessage` stream, to distinguish between aborted and finished request.

## running the tests

start the service:

```shell
SERVICES=NodeEndpoint PORT=8000 pnpm --filter @restatedev/restate-e2e-services app:http1
```

then register the http1.1 deployment

```shell
curl -X POST http://localhost:9070/deployments \
  -H 'content-type: application/json' \
  -d '{"uri": "http://localhost:8000", "use_http_11": true}'
```

and then run the test

```shell
RESTATE_INGRESS_URL=http://localhost:8080 \
RESTATE_ADMIN_URL=http://localhost:9070 \
pnpm --filter @restatedev/restate-e2e-services exec vitest run test/node_endpoint.test.ts
```

### DIFF
--- a/packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
+++ b/packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
@@ -183,7 +183,9 @@ function nodeHandlerImpl(
     // Abort controller used to cleanup resources at the end of this stream lifecycle
     const abortController = new AbortController();
     httpRequest.on("close", () => {
-      abortController.abort();
+      if (!httpRequest.readableEnded) {
+        abortController.abort();
+      }
     });
 
     const writeHead = res.writeHead.bind(res);
--- a/.github/workflows/integration-http1.yaml
+++ b/.github/workflows/integration-http1.yaml
@@ -1,0 +1,145 @@
+name: Integration HTTP/1.1
+
+# Controls when the workflow will run
+on:
+  pull_request:
+  push:
+    branches: [main, release-**]
+  schedule:
+    - cron: "0 */6 * * *" # Every 6 hours
+  workflow_dispatch:
+    inputs:
+      restateCommit:
+        description: "restate commit"
+        required: false
+        default: ""
+        type: string
+      restateImage:
+        description: "restate image, superseded by restate commit"
+        required: false
+        default: "ghcr.io/restatedev/restate:main"
+        type: string
+      serviceImage:
+        description: "service image, if provided it will skip building the image from sdk main branch"
+        required: false
+        default: ""
+        type: string
+
+jobs:
+  sdk-test-suite:
+    if: github.repository_owner == 'restatedev'
+    runs-on: warp-ubuntu-latest-x64-4x
+    name: Node HTTP/1.1 integration test
+    permissions:
+      contents: read
+      issues: read
+      checks: write
+      pull-requests: write
+      actions: read
+    strategy:
+      matrix:
+        node-version: [22.x]
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: restatedev/sdk-typescript
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          registry-url: "https://registry.npmjs.org"
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 11.1.3
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm build
+
+      # support importing oci-format restate.tar
+      - name: Set up Docker containerd snapshotter
+        uses: docker/setup-docker-action@v4
+        with:
+          version: "v28.5.2"
+          set-host: true
+          daemon-config: |
+            {
+              "features": {
+                "containerd-snapshotter": true
+              }
+            }
+
+      ### Download the Restate container image, if needed
+      # Due to https://github.com/actions/upload-artifact/issues/53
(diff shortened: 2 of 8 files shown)


aPoCoMiLogin:
@slinkydeveloper gentle ping, let me know if there's anything else needed from my side to get this merged

packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
```diff
@@ -183,7 +183,9 @@ function nodeHandlerImpl(
     // Abort controller used to cleanup resources at the end of this stream lifecycle
     const abortController = new AbortController();
     httpRequest.on("close", () => {
-      abortController.abort();
+      if (!httpRequest.readableEnded) {
```
slinkydeveloper, line 186: i'm trying to reason about this piece of code... I want the controller abort to fire on every stream closed, and listening on `httpRequest.close` is the right thing to do from what i can see in the node docs...

Which node version are you using BTW?

aPoCoMiLogin: node: 24.19.0 but that will occur in node 16+ (including latest version)

the issue is that close in the `IncomingMessage` is always triggered, no matter if it was finished correctly or aborted: https://nodejs.org/api/http.html#event-close-3

image


sorry for not explaining it thoroughly in the PR description..

slinkydeveloper: Yes but that's exactly the behavior we want: the abort controller should fire on either a normal closure or abort. That's to make sure eventual child tasks get killed if they remain dangling.

This is the related docs: https://restatedev.github.io/sdk-typescript/interfaces/_restatedev_restate-sdk.Request.html#attemptcompletedsignal

aPoCoMiLogin: I don't think this is the same, as without that fix, the handler despite finishing without errors, always returns `RestateError: (500) Connection closed` which triggers retry-policy.

```ts
import { setTimeout } from "node:timers/promises";
import * as restate from "@restatedev/restate-sdk";

async function handler(ctx: restate.Context) {
  await setTimeout(1000);
}
```

this will correctly work for http2, but it will trigger retry and `connection closed` error for http1.1

k-jarzyna:
+1

packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
nikrooz: Thanks @aPoCoMiLogin for looking into this. I think you identified the issue correctly, but I’m not sure the proposed fix fully handles it without breaking other flows.

As @slinkydeveloper mentioned The abort signal should fire when the whole attempt ends, successfully or unsuccessfully, so any remaining work can stop and clean up. this is the contract we cannot break.

Based on a quick look at this, h1 and h2 have different event lifecycles (needs some more investigation and testing)

h1:
```
Runtime finishes request body
        ↓
req emits "end"
        ↓
req emits "close"       ← only the request is complete
        ↓
handler continues
        ↓
res finishes or closes  ← the whole attempt is complete
```

h2:
```
Runtime finishes request body
        ↓
req emits "end"
        ↓
readableEnded becomes true
        ↓
handler finishes and response ends
        ↓
req emits "close"
res emits "finish"      ← same turn, whole attempt is complete
```

So the current `req.close` listener is correct for h2, but fires too early for h1, as you identified.

The `readableEnded` check avoids that early h1 abort, but then this can happen:

```
req closes with readableEnded true
        ↓
abort is skipped
        ↓
handler continues
        ↓
connection drops or response completes
        ↓
req cannot emit close again
        ↓
abort never fires
```

It also causes problems for h2. On normal completion `readableEnded` is already `true`. Node can also emit `req.end` before reporting a real `RST_STREAM`, so `readableEnded` can be `true` for a genuine stream failure too. In both cases the guard skips the abort.

I have a fix in mind, but I would like to investigate and test the lifecycle cases carefully so we do not break anything else. If you don’t mind, leave this with me for now and I will try to look at it next week.

aPoCoMiLogin: @nikrooz thanks for taking a look on this one, and yes if you have something in mind that would fix that behavior then you can take it from here, as i'm not very familiar with the underlyings of the lifecycle restate expects. for now i'm using local patch that via pnpm-patch so there is no hurry

nikrooz:
This should have been fixed by https://github.com/restatedev/sdk-typescript/pull/787 and been released in 1.16.9. Please reopen the issue if still not resolved

aPoCoMiLogin:
@nikrooz it seems that it's resolved, thanks for the fix and your time :pray:

PR #771, opened 2026-07-29
Guard request half-close abort on readableEnded

### LINKED ISSUE #745: request/response: SDK aborts the invocation on request half-close, killing suspended handlers
In `REQUEST_RESPONSE` mode the runtime half-closes the request stream once input is delivered and holds the response open for the reply. The node endpoint (`src/endpoint/node_endpoint.ts`) treats the request's `close` as fatal and aborts unconditionally, so that half-close aborts any handler currently suspended on a Call/promise. It surfaces as `[500 Internal] Connection closed` (SDK: `Error when processing a Restate context operation. RestateError: (500) Connection closed`) and a retry loop. Handlers that reply within the tick before the abort are unaffected.

```diff
   httpRequest.on("close", () => {
-    abortController.abort();
+    if (!httpRequest.readableEnded) abortController.abort();
   });
```

`complete` is the wrong guard: on Node 24 `Http2ServerRequest.complete` is `true` for both a clean half-close and a mid-stream RST, so it never distinguishes a real drop. `readableEnded` does.

`1.13.0` and `1.14.5`, Node 24, transport-independent.

nstandif:
Closes #745.

Depends on #774 (unrelated e2e test issue this change exposes; CI won't be green here until that merges).

In REQUEST_RESPONSE mode, the runtime half-closes the request after the journal is delivered and keeps the response open for the reply. The request's "close" event was aborting the invocation unconditionally, killing any handler suspended on a call or promise. Surfaced as `[500] Connection closed` plus a retry loop.

`complete` can't distinguish a clean half-close from a mid-stream RST on Node 24 (both read true). `readableEnded` can: it's only true after the stream actually saw "end".

Fix: gate the abort on `!request.readableEnded`, extracted into `abortSignalForRequest`.

Added `test/node_endpoint.test.ts` against a real `node:stream` Readable. Confirmed the test fails without the guard before adding it. Suite and typecheck pass.

No public API change, so no changeset. Flag if you want a release note anyway.

### DIFF
--- a/packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
+++ b/packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
@@ -191,11 +191,6 @@ function nodeHandlerImpl(
     const outputWriter = outputWriterAdapter(httpResponse);
     const res = httpResponse as NodeWritableResponse;
 
-    // Abort controller used to cleanup resources at the end of this stream lifecycle
-    const abortSignal = abortSignalForRequest(httpRequest);
-
-    const writeHead = res.writeHead.bind(res);
-
     // handle should never throw
     const restateResponse = handler.handle({
       url,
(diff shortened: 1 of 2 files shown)


nikrooz:
This should have been fixed by https://github.com/restatedev/sdk-typescript/pull/787 and been released in 1.16.9. Please reopen the issue if still not resolved

PR #787, opened 2026-08-24
Fix HTTP/1 attempt abort timing

nikrooz:
In HTTP/1, request close fires after Node reads the request body, while the handler may still be running. This aborted the attempt too early.
Use response close for HTTP/1. Keep request close for HTTP/2, where it marks the whole stream closing.
Adds lifecycle tests for clean completion, dropped connections, keep-alive, stream resets, and multiplexed HTTP/2 requests.

### DIFF
--- a/packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
+++ b/packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
@@ -183,6 +183,15 @@ function nodeHandlerImpl(
     // Abort controller used to clean up resources at the end of this attempt.
     const abortController = new AbortController();
     const abort = () => abortController.abort();
+
+    // HTTP/1 has separate request and response lifecycles:
+    // request body ends -> request closes -> handler finishes -> response closes.
+    // The request can therefore close while the attempt is still running. The
+    // response closes when the response completes or the connection is lost.
+    //
+    // In HTTP/2, the request and response share one stream:
+    // request body ends -> handler finishes -> stream closes -> request closes.
+    // Request close therefore covers both stream completion and reset.
     if (httpRequest.httpVersionMajor >= 2) {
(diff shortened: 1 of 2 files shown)


packages/libs/restate-sdk/src/endpoint/node_endpoint.ts
```diff
@@ -180,11 +180,23 @@ function nodeHandlerImpl(
...
+    const abort = () => abortController.abort();
+
+    // HTTP/1 has separate request and response lifecycles:
+    // request body ends -> request closes -> handler finishes -> response closes.
+    // The request can therefore close while the attempt is still running. The
+    // response closes when the response completes or the connection is lost.
+    //
+    // In HTTP/2, the request and response share one stream:
+    // request body ends -> handler finishes -> stream closes -> request closes.
+    // Request close therefore covers both stream completion and reset.
```
slinkydeveloper, line 196:
```suggestion
    // WARNING: DON'T TOUCH THIS UNLESS YOU REALLY KNOW WHAT YOU'RE DOING
    //
    // HTTP/1 has separate request and response lifecycles:
    // request body ends -> request closes -> handler finishes -> response closes.
    // The request can therefore close while the attempt is still running. The
    // response closes when the response completes or the connection is lost.
    //
    // In HTTP/2, the request and response share one stream:
    // request body ends -> handler finishes -> stream closes -> request closes.
    // Request close therefore covers both stream completion and reset.
```

slinkydeveloper (changes requested):
lgtm, just left one little comment

slinkydeveloper:
lets merge and release a patch fix :)

slinkydeveloper (approved):

slinkydeveloper:
@nikrooz can you go over prs and issues to see what we can close with this now?

Issue #566, opened 2025-07-02
Improve API around state access


slinkydeveloper:
Leading idea at the moment

```typescript
const myState = restate.state<{ a: number, b?: number }>({ a: 0 })

export const queue = object({
  name: "queue",
  states: queueStates,
  handlers: {
    done: async (ctx: ObjectContext<QueueState>): Promise<void> => {
      const a: number = await ctx.get("a") // Returns default
      const b: number | undefined = await ctx.get("b") // Returns undefined (or null?!), if not available
```

cc @nikrooz

marcus-sa:
Throwing this into the mix https://github.com/marcus-sa/deepkit-restate 😉

PR #793, opened 2026-09-02
Allow Typed State when implementing an iface

### LINKED ISSUE #791: iface implementations cannot use Typed State
Version: v1.17.0

Unable to specify a Typed State generic on `restate.ObjectContext` when implementing an iface, it currently enforces all handlers to use untyped context

Minimal repro: (this produces a TS error on `.implement`)
```ts
import * as restate from '@restatedev/restate-sdk'

const fooIface = restate.iface.object('foo', {
  getFoo: restate.iface.json<void, string | null>(),
})

export const foo = restate.implement(fooIface, {
  handlers: {
    getFoo: (ctx: restate.ObjectContext<{ foo: string }>) => ctx.get('foo'),
  },
})
```

mupperton:
Fixes #791

Usage: (UPDATED to reflect latest PR changes)
```ts
import * as restate from '@restatedev/restate-sdk'

const fooIface = restate.iface.object('foo', {
  getFoo: restate.iface.json<void, string | null>(),
})

export const foo = restate.implement(fooIface, {
  handlers: {
    getFoo: async (ctx) => {
      const value = await ctx.get('foo')
      // Correctly inferred as `string | null`
      return value
    },
  },
  states: restate.state<{ foo: string }>(),
})
```

### DIFF
--- a/packages/libs/restate-sdk/src/common_api.ts
+++ b/packages/libs/restate-sdk/src/common_api.ts
@@ -123,7 +123,7 @@ export type {
   InferInput,
   InferOutput,
 } from "@restatedev/restate-sdk-core";
-export { implement } from "./types/interface.js";
+export { implement, typed } from "./types/interface.js";
 export type {
   /** @internal */
   FnOf,
--- a/packages/libs/restate-sdk/src/common_api.ts
+++ b/packages/libs/restate-sdk/src/common_api.ts
@@ -123,7 +123,7 @@ export type {
   InferInput,
   InferOutput,
 } from "@restatedev/restate-sdk-core";
-export { implement, typed } from "./types/interface.js";
+export { implement, state } from "./types/interface.js";
 export type {
   /** @internal */
   FnOf,
--- a/packages/libs/restate-sdk/src/types/interface.ts
+++ b/packages/libs/restate-sdk/src/types/interface.ts
@@ -16,6 +16,8 @@ import type {
   ObjectSharedContext,
   WorkflowContext,
   WorkflowSharedContext,
+  TypedState,
+  UntypedState,
 } from "../context.js";
 import type {
   ServiceDefinition,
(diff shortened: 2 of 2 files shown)


slinkydeveloper:
Hi @mupperton, i would like to think about this... I was hoping we would be able to make the state support a bit more powerful, with things like "defaults": https://github.com/restatedev/sdk-typescript/issues/566

mupperton:
> Hi @mupperton, i would like to think about this... I was hoping we would be able to make the state support a bit more powerful, with things like "defaults": #566

Yeah sounds good - was hoping something could also be done for a normal VO definition (not iface + implement) as that has a _similar_ issue where the context type needs manually specifying in each handler as-is

slinkydeveloper:
would you be up to give it a shot at sorting out a solution @mupperton ? even if just for iface + implement, it would already be good enough :)

mupperton:
I could refactor this PR into setting a `states` property instead of `options.typedState`, but without the default value, as that would need the full concrete implementation within the VO and Workflow RPC methods etc too

So this would get the ball rolling at least, and open the door for a "defaults" implementation after, and availability on normal VO/WF definitions

mupperton:
Thinking about the "defaults" a bit more with that proposed API, worth noting that there is a chance of non-determinism if the invocation is paused and resumed on a different deployment which has a different default value

slinkydeveloper:
> Thinking about the "defaults" a bit more with that proposed API, worth noting that there is a chance of non-determinism if the invocation is paused and resumed on a different deployment which has a different default value

Because it is a code change, like any other code change, you should register a new revision in that case...

mupperton:
> > Thinking about the "defaults" a bit more with that proposed API, worth noting that there is a chance of non-determinism if the invocation is paused and resumed on a different deployment which has a different default value
> 
> Because it is a code change, like any other code change, you should register a new revision in that case...

Yes exactly, but the UI allows for resuming an invocation with the current journal state on a different deployment (new revision) - and if that new revision had changed the default state value, and if that particular invocation started with no existing state value, then the non-determinism could occur (depending on usage of the new default state value)

Simply raising awareness of this scenario, and that it might be worth making a note in the docs/jsdoc if/when this gets implemented with the proposed API

mupperton:
@slinkydeveloper - anything still outstanding for this?

slinkydeveloper:
@nikrooz could you help with this one, and see how we can extend to have default values here nicely?

Also, I was wondering how to have a similar functionality in the gen sdk...

slinkydeveloper:
I'm thinking how the API could look like, and was wondering if we should instead strive to align with the iface API, e.g.

```
const counterState = state({
  count: state.value<number>({ default: 0 }),
});

export const counter = object({
  name: "counter",
  handlers: {
    ...
  }.
  state: counterState
});
```

this would be more aligned with iface, and would give us room to add "state" options (e.g. which keys are eager or lazy, or the defaults, or potentially in the future the schemas too)

slinkydeveloper:
I did a prototype here: https://github.com/slinkydeveloper/sdk-typescript/tree/prototype/state-iface-dsl

PR #794, opened 2026-09-03
Fix output type of JsonSerde

### LINKED ISSUE #792: iface using JsonSerde in iface.serdes changes the expected type
Version: v1.17.0

Given a scenario of a pre-v1.17 handler with a binary input serde but using default JSON output serde that looks like this:
```ts
import * as restate from '@restatedev/restate-sdk'

export const foo = restate.service({
  name: 'foo',
  handlers: {
    // Return type of `Promise<boolean>` is inferred
    upload: restate.createServiceHandler({
      input: restate.serde.binary,
    }, async (ctx, u8) => {
      return true
    }),
  },
})
```
TS correctly infers the output type of the handler as `Promise<boolean>`

When using ifaces, the only option to represent the binary input serde is through `iface.serdes`, which ends up like this:
```ts
import * as restate from '@restatedev/restate-sdk'

const fooIface = restate.iface.service('foo', {
  upload: restate.iface.serdes({
    input: restate.serde.binary,
  }),
})

export const foo = restate.implement(fooIface, {
  handlers: {
    // Return type of `Promise<any>` is inferred
    upload: async (ctx, u8) => {
      return true
    },
  },
})
```
Which changes the inferred output type to be `Promise<any>`

So naturally next step is to specify the output serde:
```ts
import * as restate from '@restatedev/restate-sdk'

const fooIface = restate.iface.service('foo', {
  upload: restate.iface.serdes({
    input: restate.serde.binary,
    output: new restate.serde.JsonSerde<boolean>(), // Or even `restate.serde.json as restate.serde.JsonSerde<boolean>`
  }),
})

export const foo = restate.implement(fooIface, {
  handlers: {
    // Return type of `Promise<boolean | undefined>` is inferred
    upload: async (ctx, u8) => {
      return true
    },
  },
})
```
but this now infers an output type of `Promise<boolean | undefined>` which I definitely did not expect, and breaks client call sites

The only solution seems that you're forced to create a standard schema serde with a compatible validation library in order to get the original `Promise<boolean>` return type

slinkydeveloper:
Fix #792

### DIFF
--- a/packages/libs/restate-sdk-core/src/serde_api.ts
+++ b/packages/libs/restate-sdk-core/src/serde_api.ts
@@ -260,7 +260,7 @@ function isStandardJSONSchemaV1(standard: StandardSchemaV1.Props): boolean {
 }
 
 export namespace serde {
-  export class JsonSerde<T> implements Serde<T | undefined> {
+  export class JsonSerde<T> implements Serde<T> {
     contentType = "application/json";
 
     constructor(readonly jsonSchema?: object) {}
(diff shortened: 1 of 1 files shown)


slinkydeveloper:
cc @mupperton

mupperton:
Looks good, thanks

PR #795, opened 2026-09-14
Accept forced-pause semantics in pause-during-run hooks tests

AhmedSoliman:
Restate 1.8 runs invocations on vqueues by default, where an Admin API
pause aborts the running attempt immediately instead of waiting for a
suspension point. The SDK then observes the connection closing (e.g.
"Premature close") and the error hook can fire before the local run
step finishes.

Relax the pause-during-run assertions to accept both the pre-1.8
graceful sequence and the 1.8+ forced-abort sequence: fixed prefix,
remaining events in any order, either error message. The paused
invocation outcome remains the authoritative assertion.

A proposal for graceful manual pause command is detailed in https://github.com/restatedev/restate/issues/5321


tillrohrmann (approved):
LGTM. +1 for merging.

